PR #5, opened 2022-10-05
Process opening refactor

chicoxyzzy:


src/desktop/main.cc
```diff
@@ -500,61 +535,61 @@ MAIN {
...
     }
 
-    if (cmd.name == "inspect") {
+    if (message.name == "inspect") {
       window->showInspector();
       return;
     }
 
-    if (cmd.name == "background") {
+    if (message.name == "background") {
```
heapwolf, line 548: we could make this message name less ambiguous while we're here

heapwolf (approved):

src/desktop/main.cc
chicoxyzzy: There was two different `cmd` variables:
- `<os>_cmd` string
- command from webvew process or backend process
I think that message is a good name for this IPC thing, but I'm open to other alternatives

chicoxyzzy:
CI is complaining on nodejs desktop example from `socket-sdk-examples` this is intended as bootstrapping has changed order of loading. I can update examples repo and run CI agains new branch. We will need to merge examples update and then change branch back to master and merge this PR

src/desktop/main.cc
chicoxyzzy: I like `ipcMessage` as well

jwerle (approved):

chicoxyzzy:
the blocker is https://github.com/socketsupply/socket-sdk-examples/pull/36 and changing examples repo back to master

PR #9, opened 2022-10-07
fix(install): Fix Platform-Specific Build Tool Detection

numonium:
# 👋  Check It Out!

On a fresh install of `socket` on macOS / Darwin (12.6), the dependency check does not properly require `automake` (rather, `make`).

Without `automake`, `install.sh` proceeds to call `./configure`, which promptly fails due to the lack of required deps.

Accordingly, this PR **properly checks for `automake` on macOS / Darwin** and dies if not present.

Similarly, as `autoconf` is the Linux equivalent, this PR moves its respective check to the Linux-specific block.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -59,12 +59,11 @@ function advice {
 }
 
 quiet command -v make
-die $? "not ok - missing build tools, try \"$(advice "automake")\""
-
-quiet command -v autoconf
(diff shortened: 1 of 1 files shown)


jwerle (approved):

jwerle:
hello! thanks for the PR. I can confirm this works on Linux

heapwolf (approved):

PR #11, opened 2022-10-07
rename --op-current-directory to --cwd

chicoxyzzy:


chicoxyzzy:
related PR in Serverless Studio https://github.com/socketsupply/serverless-studio/pull/912

jwerle:
eventually, we should move all of these injected flags to environment variables

jwerle (approved):

PR #12, opened 2022-10-07
Fix copying options when no build field is provided in config

chicoxyzzy:


src/cli/cli.cc
```diff
@@ -1556,12 +1556,14 @@ int main (const int argc, const char* argv[]) {
 
       fs::current_path(oldCwd);
     } else {
+      // fs::remove_all(pathResourcesRelativeToUserBuild);
```
jwerle, line 1559: why not just remove the line?

Issue #23, opened 2022-10-26
Backend process and windows management improvements


chicoxyzzy:
We need to remove automatic `process.open` when we create a new window and call it manually (for the main window in most cases).

Required changes:
- [x] remove `process.open` IPC call from the window runtime preload
- [ ] add `io.backendProcess.isOpen(): boolean` and related method in Socket Runtime
- [x] add `io.backendProcess.open(force?: boolean): void` and update related method in Socket Runtime. `force` will kill old process and open new one
- [x] add `io.backendProcess.kill(): void`

We also need to improve windows management from `io`
- [x] add `io.runtime.navigate({ window, value }: { window: number, value: string }): void`

chicoxyzzy:
I have `process.isOpen()` always returning true when I check for `nullptr`. Will leave it open for now.

chicoxyzzy:
I never need this and it could be controlled from the UI. Let's close for now and re-open if there will be any use-cases

PR #28, opened 2022-11-01
Percent changes for initial width and height values

chicoxyzzy:
- [x] default value of 100% for width and height
- [x] percent window dimensions in config on mac
- [x] percent window dimensions in config on linux
- [x] percent window dimensions in config on windows
- ~~default title in config on mac~~
- ~~default title in config on linux~~
- ~~default title in config on windows~~

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -129,6 +129,7 @@ int runApp (const fs::path& path, const String& args, bool headless) {
     }
 
     status = std::system((headlessCommand + prefix + cmd + " " + args + " --from-ssc").c_str());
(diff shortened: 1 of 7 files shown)


chicoxyzzy:
I will remove title changes from here.

> sometimes we want the title of new window be different. this will lead to the bad UX (blinking). We need to add title to the navigate method instead probably

chicoxyzzy:
That also means that we need a method to `navigate` from the render process

chicoxyzzy:
a bit related issue https://github.com/socketsupply/socket/issues/23

jwerle (approved):

PR #31, opened 2022-11-15
Windows management for the render process

chicoxyzzy:
- make all windows management messages resolve in a proper window (the one they was called from, see [io changes](https://github.com/socketsupply/io/pull/44))
- allow broadcasting to all the windows via `ssc-node`'s send
- remove `getConfig` message handler from the render process (all necessary config fields are available via `window.__args` or `io.runtime.args`)


src/desktop/main.cc
```diff
@@ -353,6 +375,29 @@ MAIN {
...
+          auto window = windowFactory.getWindow(message.index);
+          if (window) {
+            window->eval(getEmitToRenderProcessJavaScript(
+              decodeURIComponent(message.get("event")),
+              value
+            ));
+          }
+        } else {
+          for (auto w : windowFactory.windows) {
+            if (w != nullptr) {
```
jwerle, line 389: `w` is a `Window*` so no need to call `getWindow()` below as `window == w` here

chicoxyzzy: > `w` is a `Window*` so no need to call `getWindow()` below as `window == w` here

it was `WindowWithMetadate*`

src/core/android.cc
```diff
@@ -233,10 +233,7 @@ jboolean NativeCore::ConfigureEnvironment () {
     }
   }
 
-  windowOptions.executable = this->config["executable"];
   windowOptions.headless = this->config["headless"] == "true";
-  windowOptions.version = "v" + this->config["version"];
-  windowOptions.title = this->config["title"];
```
chicoxyzzy, line 239: this 3 are in the config already

src/desktop/main.cc
```diff
@@ -433,12 +476,12 @@ MAIN {
           indexMain = std::stoi(message.get("indexMain"));
           indexSub = std::stoi(message.get("indexSub"));
         } catch (...) {
-          window->resolvePromise(seq, OK_STATE, "");
```
chicoxyzzy, line 436: this was failing to parse

src/desktop/main.cc
```diff
@@ -674,15 +777,14 @@ MAIN {
       return;
     }
 
-    if (message.name == "getConfig") {
-      const auto seq = message.get("seq");
-      auto wrapped = ("\"" + String(_settings) + "\"");
-      window->resolvePromise(seq, OK_STATE, encodeURIComponent(wrapped));
-      return;
-    }
```
chicoxyzzy, line 682: we have config in the `window.__args.config` / `io.runtime.args`

jwerle (approved):

PR #35, opened 2022-11-28
Cleanup initial config

chicoxyzzy:


src/cli/templates.hh
```diff
@@ -1542,13 +1539,10 @@ width: 1024
...
 
-# TODO description & value
-# mac_sign:
-
-# TODO description & value
+# macOS codesign certificate name.
 # mac_codesign_identity:
 
-# TODO description & value
+# Paths to the files that should be included in the app bundle.
```
heapwolf, line 1545: This is actually the files that need to be *signed* in the bundle, not included, that's what build is for.

src/cli/templates.hh
```diff
@@ -1542,13 +1539,10 @@ width: 1024
 # The icon to use for identifying your app on MacOS.
 # mac_icon:
 
-# TODO description & value
-# mac_sign:
-
-# TODO description & value
+# macOS codesign certificate name.
```
heapwolf, line 1542: can we have a description about what it should be? where does it come from? etc.

heapwolf: or maybe this document doesnt need descriptions and the docs site should just be improved? we could maybe even kill all the comments and just link to the docs?

src/cli/templates.hh
```diff
@@ -1432,62 +1432,59 @@ constexpr auto gDefaultConfig = R"CONFIG(
...
-version: 0.0.1
+# Advanced Compiler Settings for debug purposes (ie C++ compiler -g, etc).
+debug_flags: -g
 
-# The initial width of the first window.
-width: 1024
+# Set the limit of files that can be opened by your process.
+file_limit: 1024
+
+# An array of environment variables, separated by commas.
```
heapwolf, line 1486: variables that will be observed when your binary is run I think

heapwolf (changes requested):

PR #36, opened 2022-12-01
Add methods to get window props

chicoxyzzy:


src/desktop/main.cc
```diff
@@ -610,6 +610,59 @@ MAIN {
...
+      return;
+    }
+
+    if (message.name == "window") {
+      const auto index = message.index;
+      auto window = windowFactory.getWindow(index);
+      if (window) {
+        const auto size = window->getSize();
+        const auto title = window->getTitle();
+        SSC::String value(
```
jwerle, line 642: this is a good candidate for `SSC::JSON`:

```c++
JSON::Object value = JSON::Object::Entries {
  {"index", index},
  {"title", title},
  {"width", size.width},
  {"height", size.height}
};
```

chicoxyzzy:
`getTitle` and `getSize` only implemented for macOS

src/desktop/main.cc
```diff
@@ -610,13 +613,140 @@ MAIN {
...
+    if (message.name == "getWindows") {
+      const auto index = message.index;
+      const auto showTitle = message.get("title") == "true" ? true : false;
+      const auto showSize = message.get("size") == "true" ? true : false;
+      const auto showStatus = message.get("status") == "true" ? true : false;
+      const auto window = windowManager.getWindow(index);
+      StringStream ss;
+      bool first = true;
+      ss << "[";
+      for (auto windowWithMetadata : windowManager.windows) {
```
jwerle, line 625: maybe we can add a `json()` method to the `ManagedWindow` class that returns a `JSON::Object` and a `json()` to `WindowManager` that returns an `JSON::Array` of those objects

chicoxyzzy: that's a good idea!

chicoxyzzy: @jwerle how can I implement these conditions using `JSON::Object`?

```js
            if (showTitle) {
               const auto title = w->getTitle();
               ss << "\"title\":\"" << title << "\",";
             }
             if (showSize) {
               const auto size = w->getSize();
               ss << "\"width\":" << size.width << ",";
               ss << "\"height\":" << size.height << ",";
             }
             if (showStatus) {
               const auto status = windowWithMetadata->status;
               ss << "\"status\":" << status << ",";
             }
             ss << "\"index\":" << windowWithMetadata->opts.index;
             ss << "}";
```

chicoxyzzy: aha, Object::Entries is `std::map`!

chicoxyzzy:
@jwerle @heapwolf I think it's ready. I can't run apps on Linux and Windows, so I need help with `getTitle` and `getSize` on these OSs

src/desktop/main.cc
chicoxyzzy: done

src/window/apple.mm
```diff
@@ -732,6 +732,10 @@ - (NSString*) filePromiseProvider: (NSFilePromiseProvider*)filePromiseProvider f
     }
   }
 
+  SSC::String Window::getTitle () {
+    return SSC::String([this->window.title UTF8String]);
+  }
+
   void Window::setTitle (const SSC::String& seq, const SSC::String& value) {
```
jwerle, line 739: We could also store the title on the `Window` or `ManagedWindow` class instance too

src/window/linux.cc
```diff
@@ -675,6 +675,10 @@ namespace SSC {
     }
   }
 
+  SSC::String Window::getTitle () {
+    // TODO: implement
```
jwerle, line 679: `return String(gtk_window_get_title(GTK_WINDOW(window)));`

src/window/linux.cc
```diff
@@ -744,6 +748,10 @@ namespace SSC {
     gtk_dialog_run(GTK_DIALOG(dialog));
   }
 
+  ScreenSize Window::getSize () {
+    // TODO: implement
```
jwerle, line 752:
```c
Screen size size = {0};
gtk_window_get_size(GTK_WINDOW(window), &size.width, &size.height);
return size;
```

src/window/win.cc
```diff
@@ -1036,6 +1036,10 @@ namespace SSC {
     });
   }
 
+  SSC::string Window::getTitle () {
+    // TODO: implement.
```
jwerle, line 1040: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowtexta

jwerle:
```c++
char title[1024] = {0};
auto size = GetWindowText(window, title, 1024);
return String(title, size);
```

jwerle (approved):
LGTM. The missing implementations are pretty straightforward. We can follow up in a subsequent PR if you don't want to add them here. I suggest we track the title of the window for fast read access instead of calling these platform APIs each time

src/window/apple.mm
heapwolf: is there an opportunity for it to become out of sync though if we do that? this isn't in the performance path so I'd say its fine.

PR #37, opened 2022-12-05
send from render

chicoxyzzy:


src/desktop/main.cc
```diff
@@ -582,6 +582,28 @@ MAIN {
       return;
     }
 
+    if (message.name == "send") {
+      const auto event = decodeURIComponent(message.get("event"));
+      const auto value = decodeURIComponent(message.get("value"));
+      const auto targetWindowIndex = message.get("window").size() >= 0 ? std::stoi(message.get("window")) : -1;
```
jwerle, line 588: We use `index` everywhere else

jwerle: oh nvm we are using it for the source window!

chicoxyzzy:
Converting this back to draft until I add a method to send IPC messages to backend process and create a module for that. That module should also make [`ssc-node`](https://github.com/socketsupply/ssc-node) obsolete and simplify writing tests

jwerle (approved):

PR #39, opened 2022-12-06
add --stdin to read from stdin and emit to window 0 as an event

heapwolf:
Pipe only 😎

### DIFF
--- a/src/desktop/main.cc
+++ b/src/desktop/main.cc
@@ -86,6 +86,7 @@ MAIN {
   SSC::StringStream argvForward;
 
   bool isCommandMode = false;
+  bool isReadingStdin = false;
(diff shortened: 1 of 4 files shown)


src/desktop/main.cc
```diff
@@ -945,6 +955,26 @@ MAIN {
...
+  if (isReadingStdin || isReadingTTY) {
+    std::string value;
+    std::thread t([&]() {
+      do {
+        if (value.size() == 0) {
+          std::getline(std::cin, value);
+        }
+
+        if (value.size() > 0) {
+          std::this_thread::sleep_for(std::chrono::milliseconds(256));
```
jwerle, line 967: 256 seems pretty high for read latency

PR #40, opened 2022-12-07
Android IPC Refactor

jwerle:


src/core/json.cc
```diff
@@ -5,6 +5,72 @@ namespace SSC::JSON {
...
+      stream << value;
+
+      if (--count > 0) {
+        stream << std::string(",");
+      }
+    }
+
+    stream << std::string("}");
+    return stream.str();
+  }
```
chicoxyzzy, line 51: hmm... I think this was working already?

jwerle: yeah it was - I just moved the implementation into `json.cc`

chicoxyzzy (commented):
I think we can rename `src/mobile/` to `src/ios` now

jwerle:
> I think we can rename `src/mobile/` to `src/ios` now

Yeah it would be nice to also clean up and consolidate the iOS stuff too

chicoxyzzy (commented):
if we drop `realpath` everywhere, we also can drop installing `coreutils` in https://github.com/socketsupply/socket/blob/ipc-cleanup/.github/workflows/ci.yml (on lines 208 and 255)

jwerle:
> if we drop `realpath` everywhere, we also can drop installing `coreutils` in https://github.com/socketsupply/socket/blob/ipc-cleanup/.github/workflows/ci.yml (on lines 208 and 255)

dropped 🤓

PR #42, opened 2022-12-07
fix: use -O0 and -g in `bin/install.sh` when DEBUG=1 is used

aleclarson:
This allows debuggable builds of the Socket runtime.

```sh
DEBUG=1 bin/install.sh
```

This replaces the `-Os` flag with `-O0 -g` flags.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -91,7 +91,7 @@ function _build_cli {
   local output_directory="$BUILD_DIR/$arch-$platform"
 
   local ldflags=($("$root/bin/ldflags.sh" --arch "$arch" --platform "$platform" -l{uv,socket-runtime}))
-  local cflags=($("$root/bin/cflags.sh" -Os))
+  local cflags=($("$root/bin/cflags.sh"))
 
   local sources=($(find "$src"/cli/*.cc 2>/dev/null))
   local outputs=()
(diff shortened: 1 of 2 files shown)


bin/install.sh
```diff
@@ -398,8 +398,12 @@ function _compile_libuv {
 
 function _check_compiler_features {
   echo "# checking compiler features"
-  local cflags=($("$root/bin/cflags.sh" -Os))
-  $CXX -x c++ "${cflags[@]}" -o /dev/null - >/dev/null << EOF_CC
+  local cflags=($("$root/bin/cflags.sh"))
+  if [[ -z "$DEBUG" ]]; then
+    cflags+=(-o /dev/null)
+  fi
```
aleclarson, line 404: Had to omit `-o /dev/null` when `-g` exists in cflags. 

Otherwise you get this error:
```
❯ bin/install.sh
# checking compiler features
error: cannot parse the debug map for '/dev/null': The file was not recognized as a valid object file
clang: error: dsymutil command failed with exit code 1 (use -v to see invocation)
not ok - /usr/bin/clang++ (14.0.0) clang > 11 is required for building socket - please report (https://discord.gg/YPV32gKCsH)
```

chicoxyzzy (approved):

PR #43, opened 2022-12-07
fix: call `stat` correctly on Darwin

aleclarson:
⚠️ Only tested on Darwin

https://github.com/socketsupply/socket/commit/178580d65f733c1e18387c860ab5ae5ff8135177#r92242711


jwerle (approved):
🚀

chicoxyzzy (approved):

PR #45, opened 2022-12-08
feat(ios): reload the webview on cmd+r

aleclarson:
When the web view's JS thread hits a deadlock, the Safari inspector becomes unresponsive and you have to manually close the app to restart it.

This PR adds a native `cmd+r` shortcut that replaces the web view instance, thereby avoiding the JS thread deadlock. I tried just using `[self.webview reload]` but it doesn't work in the  case of a deadlock.

Swizzling code lifted from React Native

## Usage

1. When using iOS simulator, you need to bind the `Record Screen` shortcut to something else, as its default binding is `cmd+r`. To do this, go to `System Preferences > Keyboard > Shortcuts > App Shortcuts`. You'll need to find where your Simulator app is in the filesystem.  
    
    image

2. Then make sure that `I/O > Input > Send Keyboard Input to Device` is enabled. Also make sure `I/O > Keyboard > Connect Hardware Keyboard` is enabled too. These are found in the menubar of the Simulator app.

### DIFF
--- a/src/mobile/ios.mm
+++ b/src/mobile/ios.mm
@@ -1,3 +1,5 @@
+#include <objc/runtime.h>
+#include <objc/message.h>
 #include "../core/core.hh"
 #include "../ipc/ipc.hh"
 #include "../window/window.hh"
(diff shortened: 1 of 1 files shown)


src/mobile/ios.mm
```diff
@@ -183,6 +198,28 @@ -           (BOOL) application: (UIApplication*) application
...
+  return YES;
+}
+
+- (void)setUpWebView
+{
+  [self.content removeAllUserScripts];
+  self.webview.navigationDelegate = nil;
+  self.webview.scrollView.delegate = nil;
+  [self.webview stopLoading];
+  [self.webview removeFromSuperview];
```
aleclarson, line 221: Not sure why, but this isn't enough for the previous web view to be released. It still shows up in Safari inspector's `Develop` menu.

aleclarson:
Closing this for now, as `isDebugEnabled` doesn't allow the private APIs used here to be removed in production builds.

I have another approach that I like better anyway. It sends a ping message to JS and waits for a pong response. Once enough time passes, a `UIAlertController` is used to ask the developer if they want to realloc the web view, as it seems to be stalled. The reason I don't just realloc the web view automatically is that the Safari inspector being paused can create a false positive (yes, the JS thread is unresponsive, but that's on purpose when debugging JS).

PR #50, opened 2022-12-12
Run io tests on CI

chicoxyzzy:
Trying to run io tests on Linux, macOS, Android Emulator and iOS Simulator


chicoxyzzy:
I'm going to merge this, we can fix mobile builds later

PR #57, opened 2022-12-15
Remove realpath dep (not available on macos)

mribbons:
I'm not sure if the intention here was to be able to compile from anywhere, but the previous code only worked from repo root anyhow, otherwise it can't compile because src isn't found.


bin/realpath.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
jwerle, line 1: I think we can drop this file since it isn't used

jwerle (changes requested):

bin/realpath.sh
mribbons: Oops should have been removed.

trevnorris (approved):

trevnorris:
Landed on the `ipc-cleanup` branch in commit 4372d3727.

PR #58, opened 2022-12-15
bin: add options to Windows build script

trevnorris:
```
* Options added:

-debug: Do a debug build of libuv and ssc. Install libuv's pdb file for
    debugging info.
-skipwebview: Skip downloading webview. User can use this if they've
    already downloaded WebView2 before and are doing a rebuild.
-webview: Pass in the version of webview to download.
-uv: Pass in the tag uv will checkout. This option will not override a
    previously checked out version of libuv. In the future it should be
    fixed to have the install script check which tag is checked out and
    do a comparison of which version is required. Then redownload and
    rebuild if necessary.

* Checkout libuv at a specific tag instead of at latest v1.x branch.

* Always force the rebuild of libuv. cmake will automatically skip the
  build if it's already been done in the past.

* Force copy all the files every time on install. In case some of the
  build options have changed since the last install.

* Removed some commented code that's no longer used.
```

### DIFF
--- a/bin/install.ps1
+++ b/bin/install.ps1
@@ -1,3 +1,5 @@
+param([Switch]$debug, [Switch]$skipwebview, $webview = "1.0.1369-prerelease", $uv = "v1.44.2")
+
 $OLD_CWD = (Get-Location).Path
 
 $ASSET_PATH = "$env:LOCALAPPDATA\Programs\socketsupply"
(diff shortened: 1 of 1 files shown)


bin/install.ps1
```diff
@@ -1,3 +1,5 @@
+param([Switch]$debug, [Switch]$skipwebview, $webview = "1.0.1369-prerelease", $uv = "v1.44.2")
```
jwerle, line 1: 🔥

jwerle (approved):

PR #59, opened 2022-12-16
typed config

heapwolf:
This makes `ssc.conf` an actual program. So now its type safe. There are some other obvious upsides and escape hatches. The only requirement of the program is that is produces a variable called "config".

```c++
Config config {
  .build = "node bin/build.js",
  .bundle_identifier = "co.socketsupply.simulator",
  .copyright = "(c) Socket Supply Co.",
  .description = "A UI for visualizing Stream Relay Protocol simulations",
  .env = "USER, TMPDIR, PWD",
  .executable = "srv",
  .flags = "-O3",
  .debug = {
    .flags = "-g"
  },
  .file_limit = 1024,
  .window = {
    .height = 750,
    .width = 1024,
  },
  .lang = "en-us",
  .maintainer = "Beep Boop Corp.",
  .name = "srv",
  .output = "dist",
  .version = "0.0.1"
};
```

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -370,6 +370,71 @@ static String getCxxFlags() {
   return flags.size() > 0 ? " " + flags : "";
 }
 
+static Map getConfig (fs::path p) {
+  std::stringstream compileConfigCommand;
+
+  if (getEnv("CXX").size() == 0) {
+    log("warning! $CXX env var not set, assuming defaults");
+
+    if (platform.win) {
+      setEnv("CXX=clang++");
+    } else {
+      setEnv("CXX=/usr/bin/clang++");
+    }
+  }
+
+  compileConfigCommand
+    << getEnv("CXX")
+    << " -std=c++2a -v"
+    << " -I" << prefixFile()
+    << " -I" << prefixFile("src/config.hh")
+    << " -I" << p
+    << " -I" << fs::current_path()
+    << " " << prefixFile("src/config.cc")
+    << " -o " << prefixFile("bin/ssc-config")
+  ;
+
+  auto r = exec(compileConfigCommand.str());
+
+  if (r.exitCode != 0) {
+    log("Unable to compile the config file");
+    log(r.output);
+    exit(r.exitCode);
+  }
+
+  // serialize and write the config to p
+  auto base = p.remove_filename().string();
+  auto cmd = prefixFile("bin/ssc-config") + " " + base;
+  r = exec(cmd);
+
+  if (r.exitCode != 0) {
+    log("Unable to serialize and write the config file to disk");
+    log(r.output);
+    exit(r.exitCode);
+  }
+
+  auto file = fs::path { p / ".ssc.dat" };
+  std::ifstream rs(file, std::ios::out | std::ios::binary);
+
+  if (!rs) {
+    log("unable to open .ssc.dat");
+    exit(1);
+  }
+
+  Config config;
+  rs.read((char*) &config, sizeof(Config));
+  rs.close();
+
+  if (!rs.good()) {
+    log("Unable to read .ssc.dat");
+    exit(1);
+  }
+
+  fs::remove(file);
+
+  return configToMap(config);
+}
+
 void printHelp (const String& command) {
   if (command == "ssc") {
     std::cout << tmpl(gHelpText, defaultTemplateAttrs) << std::endl;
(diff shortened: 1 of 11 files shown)


src/cli/cli.cc
```diff
@@ -370,6 +370,71 @@ static String getCxxFlags() {
...
+    } else {
+      setEnv("CXX=/usr/bin/clang++");
+    }
+  }
+
+  compileConfigCommand
+    << getEnv("CXX")
+    << " -std=c++2a -v"
+    << " -I" << prefixFile()
+    << " -I" << prefixFile("src/config.hh")
```
jwerle, line 390: this should just be:

```c++
<< " -I" << prefixFile("src")
```

src/cli/cli.cc
```diff
@@ -370,6 +370,71 @@ static String getCxxFlags() {
...
+    << " -I" << p
+    << " -I" << fs::current_path()
+    << " " << prefixFile("src/config.cc")
+    << " -o " << prefixFile("bin/ssc-config")
```
jwerle, line 394: maybe this should build to a local dot directory (`.ssc/tmp`) since the files where `prefixFile()` looks is global in the file system

chicoxyzzy:
@heapwolf `height` and `weight` could be a string when they are in percent, i.e. `"100%"`

src/desktop/main.cc
```diff
@@ -261,8 +259,8 @@ MAIN {
     return exitCode;
   }
 
-  String initialHeight = app.appData["height"].size() > 0 ? app.appData["height"] : "100%";
-  String initialWidth = app.appData["width"].size() > 0 ? app.appData["width"] : "100%";
+  String initialHeight = app.appData["window_height"].size() > 0 ? app.appData["window_height"] : "100%";
+  String initialWidth = app.appData["window_width"].size() > 0 ? app.appData["window_width"] : "100%";
```
chicoxyzzy, line 263: `"100%"` string is a default value

src/config.cc
```diff
@@ -0,0 +1,23 @@
...
+  std::filesystem::path dest = { base / ".ssc.dat"};
+
+  std::ofstream ws(dest, std::ios::out | std::ios::binary);
+  if (!ws) return 1;
+
+  using namespace SSC;
+
+  #include "ssc.conf" // NOLINT
+
+  ws.write((char*) &config, sizeof(Config));
```
jwerle, line 17: we should enclose `config` in a function:

```c++
static void configure (Config& config) {
  #include "ssc.conf" // NOLINT
}
```

The declaration of `config` should be in `main()` or static to the file so the caller need not worry about it. They'll still be able to do:

```c++
config = {
  // fields
};
```

src/desktop/main.cc
```diff
@@ -370,8 +368,9 @@ MAIN {
         return;
       }
 
-      if (message.name == "getConfig") {
-        window->resolvePromise(seq, OK_STATE, _settings);
+      if (message.name == "config") {
```
jwerle, line 371: for a friendlier API we could support JSON if `key` is not given:

```c++
if (message.name == "config") {
  if (message.has("key")) {
    auto key = message.get("key");
    window->resolvePromise(seq, OK_STATE, app.appData[key]);
  } else {
    window->resolvePromise(seq, OK_STATE, JSON::Object(app.appData).str());
  }
  return;
}
```

src/cli/templates.hh
```diff
@@ -1351,153 +1351,146 @@ constexpr auto gStoryboardLaunchScreen = R"XML(<?xml version="1.0" encoding="UTF
...
 
-# Helps to make your app searchable in Linux desktop environments.
-# linux_categories: Developer Tools
+  // A string that indicates the version of the application. It should be a semver triple like 1.0.0
+  .version = "0.0.1",
 
-# The command to execute to spawn the "back-end" process.
-# linux_cmd: beepboop
+  // Files that should be added to the compile step.
+  .files = "native-module1.cc native-module2.cc",
```
jwerle, line 1424: 👍🏼

src/desktop/main.cc
chicoxyzzy: Can we make it a `std::variant<int, String>`?

src/desktop/main.cc
chicoxyzzy: I think in needs to be wrapped in `encodeURIComponent` as well

chicoxyzzy:
Alternative solution could be a toml format, which has types. It will require a parser, but maybe it's ok for the `ssc` binary?

chicoxyzzy:
This parser has a nice `value_or(value)` feature to set defaults https://github.com/marzer/tomlplusplus

src/cli/cli.cc
heapwolf: I think we should open a separate issue for this, I like the idea. There are some other files that we may need in the future. This is probably a part of a broader conversation.

Issue #60, opened 2022-12-16
Document `ssc.config`


nichoth:
Should have documentation about the file `ssc.config` — what fields are required, etc.

chicoxyzzy:
the documentation is here https://sockets.sh/config/

it needs an update

nichoth:
> the documentation is here https://sockets.sh/config/
> 
> it needs an update

I see, in the foot part. Thanks

<img width="248" alt="image" src="https://user-images.githubusercontent.com/5776508/208145377-8460b7b9-f96f-4b78-a178-488b9e2b82df.png">

chicoxyzzy:
Also `ssc init` creates a config with comments.

There is a PR which renames it to `ssc.conf` https://github.com/socketsupply/socket/pull/59

PR #61, opened 2022-12-16
add a link to config docs

nichoth:
Link to the `ssc.config` docs directly. I had trouble finding that page on the main website.

related to #60

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ UDP and File I/O.
 
 Please read (and contribute to) the documentation [here](https://sockets.sh).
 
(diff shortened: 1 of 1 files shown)


chicoxyzzy:
> I had trouble finding that page on the main website.

Me too. Hmm.. probably we need a better consolidated docs hierarchy

chicoxyzzy:
Oh! I think we should always target `ipc-cleanup` branch

chicoxyzzy:
you can drop first for commits in this branch. they are from master branch

PR #62, opened 2022-12-17
Fix android config args

mribbons:


src/android/webview.kt
```diff
@@ -52,6 +52,10 @@ open class WebViewClient (activity: WebViewActivity) : android.webkit.WebViewCli
   ): Boolean {
     val url = request.url
 
+    if (url.scheme == "http") {
```
jwerle, line 55: maybe `https` too?

src/android/window.cc
```diff
@@ -40,6 +40,7 @@ namespace SSC::android {
     options.debug = isDebugEnabled() ? true : false;
     options.env = stream.str();
     options.cwd = rootDirectory.str();
+    options.appData = this->config;
```
jwerle, line 43: Good catch

src/cli/cli.cc
```diff
@@ -1198,6 +1198,16 @@ int main (const int argc, const char* argv[]) {
         }
       }
 
+      if (settings["android_allow_cleartext"].size() == 0)
+      {
```
jwerle, line 1202: We prefer left curly braces to be in the same line as the right paren

src/cli/cli.cc
```diff
@@ -1244,6 +1254,7 @@ int main (const int argc, const char* argv[]) {
         split(settings["android_native_sources"], ' ')
       ) {
         auto filename = fs::path(file).filename();
+        log(String("Android NDK source: " + String(targetPath / file)).c_str());
```
jwerle, line 1257: Can you remove this? We actually removed it in another commit

jwerle (changes requested):

src/android/webview.kt
mribbons: 👍

src/android/window.cc
mribbons: 🙏

src/cli/cli.cc
mribbons: 👍

src/cli/cli.cc
mribbons: 👍

chicoxyzzy (approved):
This change fixes CI build for Android

jwerle (approved):

PR #66, opened 2022-12-19
Fix crash and remove opaque ptr

trevnorris:
```
window: use unique app ptr in lambda

When shutdownHandler called from the windowManager instance (assigned as
onExit) the member variable "app" is used instead of the stack allocated
App instance.

Trying to access the member variable app after running
windowManager.destroy() causes a segfault since the reference on the
WindowManager instance is no longer valid.

To get around this assign the App instance to a static pointer variable
and use that instead.

Make the App and WindowManager instances static to simplify
initialization order and remove the possibility for a race condition.
```
```
app: don't use opaque ptr

No reason to store the WindowManager instance as an opaque pointer since
none of the member methods of WindowManager are being used. Instead
forward declare it and store the pointer directly.
```

### DIFF
--- a/src/app/app.cc
+++ b/src/app/app.cc
@@ -173,11 +173,11 @@ namespace SSC {
     return cwd;
   }
 
-  void App::setWindowManager (void *windowManager) {
+  void App::setWindowManager (WindowManager* windowManager) {
     this->windowManager = windowManager;
   }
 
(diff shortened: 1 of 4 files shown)


chicoxyzzy (approved):
can confirm this fixes the issue on macOS 👍

PR #70, opened 2022-12-22
android: Resolve windows build issues

mribbons:
bin\install.ps1 wasn't copying libuv src files
(Caused silent crash on copy files)
Note we currently require
bin\install.ps1 --uv 1.x
until 1.44.5 can be set as the default option.
Also resolved android command line tool path issues for windows, this relies on ANDROID_HOME being set. Also resolved --host option not being a valid argument


src/cli/cli.cc
```diff
@@ -895,21 +900,28 @@ int main (const int argc, const char* argv[]) {
...
-          if (r.exitCode == 0) {
-            devHost = r.output;
+      if (hostArg.size() == 0) {
+        hostArg = optionValue(arg, "--host");
+        if (hostArg.size() > 0) {
+          devHost = hostArg;
+        } else {
+          if (flagBuildForIOS || flagBuildForAndroid) {
+            auto r = exec((!platform.win) ? 
+              "ifconfig | grep -w 'inet' | awk '!match($2, \"^127.\") {print $2; exit}' | tr -d '\n'" :
```
heapwolf, line 910: this will fail on macOS because of the quote escapes in awk's match expression. removing them fixes it.

```bash
ifconfig | grep -w 'inet' | awk '!match($2, "^127.") {print $2; exit}' | tr -d '\n'
```

heapwolf (changes requested):

src/cli/cli.cc
heapwolf: Maybe we could do this instead of spawning so many external programs...

```
#include <stdlib.h>
#include <unistd.h>
#include <netdb.h>
#include <arpa/inet.h>

char* getIP () {
   char host[256];
   struct hostent *host_entry;
   gethostname(host, sizeof(host));
   host_entry = gethostbyname(host);
   return inet_ntoa(*((struct in_addr*) host_entry->h_addr_list[0]));
}
```

src/cli/cli.cc
```diff
@@ -1903,6 +1915,10 @@ int main (const int argc, const char* argv[]) {
       StringStream packages;
       StringStream gradlew;
 
+      if (settings["android_accept_sdk_licenses"].size() > 0) {
```
jwerle, line 1918: 👍

jwerle (approved):

src/cli/cli.cc
mribbons: @heapwolf this is still showing as Changes Requested even though you have resolved it.

Have had a look at replacing the shell commands with libuv as discussed in discord..
The issue is that the shell commands return useful values, because they tend to locate an address with a gateway which is most likely
to be an non internal, non virtual ip address.
I can't see any way with libuv to get an address with a gateway. The first non internal address returned by libuv tends to be just a random virtual interface, or one that isn't connected (eg vpn, ethernet on mac os, WSL on windows)
Based on these two posts, there dooesn't seem to be a clean way to do it in general or using libuv.
https://stackoverflow.com/a/3288983/3739540
https://github.com/nodejs/node/issues/5417

PR #74, opened 2022-12-24
Implement last 2 Windows APIs causing a crash.

trevnorris:
Still need to finish implementing `Window::getSize()`.

### DIFF
--- a/src/window/win.cc
+++ b/src/window/win.cc
@@ -1140,7 +1140,19 @@ namespace SSC {
   }
 
   ScreenSize Window::getSize () {
-    // TODO: implement.
+    // 100 are the min width/height that can be returned. Keep defaults in case
+    // the function call fail.
+    UINT32 height = 100;
+    UINT32 width = 100;
+    RECT rect;
+
+    // Make sure controller exists, and the call to get window bounds succeeds.
+    if (controller != nullptr && controller->get_Bounds(&rect) >= 0) {
+      height = rect.bottom - rect.top;
+      width = rect.right - rect.left;
+    }
+
+    return { static_cast<int>(height), static_cast<int>(width) };
   }
 
   void Window::setSize (const SSC::String& seq, int width, int height, int hints) {


src/window/win.cc
```diff
@@ -1120,7 +1125,11 @@ namespace SSC {
   }
 
   SSC::String Window::getTitle () {
-    // TODO: implement.
+    LPWSTR title;
+    webview->get_DocumentTitle(&title);
```
chicoxyzzy, line 1129: is it document's title or native window title? should be the latter

trevnorris: If that's the case then I'll switch this to use `GetWindowText()`.

trevnorris: Fixed.

trevnorris:
Having a problem because the `add_NewWindowRequested()` callback isn't running during the `socket-api` tests. So I can't store `ICoreWebView2NewWindowRequestedEventArgs` and use it to get `ICoreWebView2WindowFeatures`. Still working through this.

jwerle:
> Having a problem because the `add_NewWindowRequested()` callback isn't running during the `socket-api` tests. So I can't store `ICoreWebView2NewWindowRequestedEventArgs` and use it to get `ICoreWebView2WindowFeatures`. Still working through this.
> 
> Notes:
> 
> * `WebView2.h` shows that the only way to retrieve `ICoreWebView2WindowFeatures*` is via the `ICoreWebView2NewWindowRequestedEventArgs::get_WindowFeatures()` API.
> * The callback passed to `add_NewWindowRequested()` is only called when `window.open()` is called from JS.

I don't think we use `window.open()` in any of the tests nor internally in `socket-api`. I think you handle it correctly here https://github.com/socketsupply/socket/pull/74/commits/bbc034db23b1f2894b70042eb8e19c1e4ee04d26

src/window/win.cc
```diff
@@ -887,9 +887,18 @@ namespace SSC {
...
+                  // TODO(trevnorris): This callback is only called when window.open() is called,
+                  // but for some reason the callback is never being called from the socket-api
+                  // tests. Because of this we can't store requested_event_args, which means that
+                  // getSize() can't properly retrieve the width and height of the window.
                   webview->add_NewWindowRequested(
                     Microsoft::WRL::Callback<ICoreWebView2NewWindowRequestedEventHandler>(
                       [&](ICoreWebView2* wv, ICoreWebView2NewWindowRequestedEventArgs* e) {
+                        Window* w = reinterpret_cast<Window*>(GetWindowLongPtr((HWND)window, GWLP_USERDATA));
+                        if (w->requested_event_args == nullptr) {
+                          abort();
```
jwerle, line 899: Should we print a warning for this?

jwerle (approved):

src/window/win.cc
```diff
@@ -1135,7 +1150,25 @@ namespace SSC {
...
+    } else {
+      HWND handle = GetForegroundWindow();
+      RECT info;
+      if (GetWindowRect(handle, &info)) {
+        height = info.bottom - info.top;
+        width = info.right - info.left;
+      }
+    }
+
+    return { static_cast<int>(height), static_cast<int>(width) };
```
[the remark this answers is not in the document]
trevnorris: Problem is the `socket-api` tests fails using the system window size. It returns `435` instead of `400` pixels wide.

src/window/win.cc
trevnorris: Adding the `abort()` was for testing. `requested_event_args` should be `nullptr` at this point, and I wanted the program to abort so I could collect some info in VS. Problem is this callback is never called. So I could never get it to abort.

jwerle:
I think `GetForegroundWindow()` could return a window not related to the instance. It looks like we handle [`WM_SIZE`](https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-size) here https://github.com/socketsupply/socket/blob/master/src/window/win.cc#L1621-L1628 with a call to `controller->put_Bounds()`. Maybe we can use [`controller->get_Bounds()`] to get the height and width of the webview window

chicoxyzzy (approved):

PR #75, opened 2022-12-24
android: Resolve windows build issues

mribbons:
bin\install.ps1 wasn't copying libuv src files
(Caused silent crash on copy files)
Note we currently require
bin\install.ps1 --uv 1.x
until 1.44.5 can be set as the default option.
Also resolved android command line tool path issues for windows, this relies on ANDROID_HOME being set. Also resolved --host option not being a valid argument

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -796,7 +796,7 @@ int main (const int argc, const char* argv[]) {
     exit(0);
   });
 
-  createSubcommand("build", { "--platform", "--port", "--quiet", "-o", "-r", "--prod", "-p", "-c", "-s", "-e", "-n", "--test", "--headless" }, true, [&](const std::span<const char *>& options) -> void {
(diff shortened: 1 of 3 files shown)


src/cli/cli.cc
```diff
@@ -899,21 +904,28 @@ int main (const int argc, const char* argv[]) {
...
-          auto r = exec("ifconfig | grep -w 'inet' | awk '!match($2, \"^127.\") {print $2; exit}' | tr -d '\n'");
-          if (r.exitCode == 0) {
-            devHost = r.output;
+      if (hostArg.size() == 0) {
+        hostArg = optionValue(arg, "--host");
+        if (hostArg.size() > 0) {
+          devHost = hostArg;
+        } else {
+          if (flagBuildForIOS || flagBuildForAndroid) {
+            auto r = exec((!platform.win) ?
```
jwerle, line 913: @heapwolf you suggest something instead of this, right?

jwerle (commented):
LGTM

We need to QA before we can approve

heapwolf (approved):

jwerle (approved):

PR #76, opened 2022-12-24
Android manual ndk build - 4.5x build time improvement

mribbons:


src/cli/cli.cc
```diff
@@ -1246,6 +1263,27 @@ int main (const int argc, const char* argv[]) {
...
+        settings["android_defaultConfig_externalNativeBuild"].assign(
+          "    externalNativeBuild {\n"
+          "      ndkBuild {\n"
+          "        arguments \"NDK_APPLICATION_MK:=src/main/jni/Application.mk\"\n"
+          "      }\n"
+          "    }"
+        );
+
+        settings["android_android_externalNativeBuild"].assign(
+          "  externalNativeBuild {\n"
```
jwerle, line 1277: these key seems awkward and potentially error prone, maybe something like:

```
android_external_native_build
```

src/cli/cli.cc
```diff
@@ -1910,7 +1935,13 @@ int main (const int argc, const char* argv[]) {
 
       StringStream sdkmanager;
       StringStream packages;
-      StringStream gradlew;
+      StringStream gradlew;      
+      String ndkVersion = "25.0.8775105";
```
jwerle, line 1939: eventually we may support this in `socket.ini`

src/cli/cli.cc
```diff
@@ -1943,12 +1976,84 @@ int main (const int argc, const char* argv[]) {
 
       if (std::system(sdkmanager.str().c_str()) != 0) {
         log("error: failed to initialize Android SDK (sdkmanager)");
+        log("You may need to add accept_sdk_licenses = \"y\" to the [android] section of socket.ini.");
```
jwerle, line 1979: @heapwolf is there a similar flow for iOS?

src/cli/templates.hh
```diff
@@ -1008,23 +1008,15 @@ android {
       abiFilters {{android_ndk_abi_filters}}
     }
 
-    externalNativeBuild {
-      ndkBuild {
-        arguments "NDK_APPLICATION_MK:=src/main/jni/Application.mk"
-      }
-    }
+{{android_defaultConfig_externalNativeBuild}}
```
jwerle, line 1011: same for this key:

```
android_default_config_external_native_build
```

jwerle (changes requested):

src/cli/cli.cc
mribbons: @jwerle happy to change this, was trying to stay with the formula of having everything start with android_ that is in the template, but I think I got a bit meta.

src/cli/cli.cc
mribbons: Agree, all of these android strings need to be configurable.

src/cli/cli.cc
jwerle: no worries! just wanting things to be consistent 🌴

src/cli/cli.cc
```diff
@@ -1246,6 +1263,27 @@ int main (const int argc, const char* argv[]) {
...
+
+        settings["android_external_native_build"].assign(
+          "  externalNativeBuild {\n"
+          "    ndkBuild {\n"
+          "      path \"src/main/jni/Android.mk\"\n"
+          "    }\n"
+          "  }"
+        );
+      } else {
+        settings["default_config_external_native_build"].assign("    // externalNativeBuild called manually for -j parallel support. Disable with [android]...enable_standard_ndk_build = true in socket.ini\n");
```
jwerle, line 1284: oops, this should be `android_default_config_external_native_build`

src/cli/templates.hh
```diff
@@ -1008,23 +1008,15 @@ android {
       abiFilters {{android_ndk_abi_filters}}
     }
 
-    externalNativeBuild {
-      ndkBuild {
-        arguments "NDK_APPLICATION_MK:=src/main/jni/Application.mk"
-      }
-    }
+{{default_config_external_native_build}}
```
jwerle, line 1011: same here: `android_default_config_external_native_build`

jwerle (changes requested):

src/cli/cli.cc
mribbons: Ok, so the reason the other one was android_android is because it's related to the android settings, but it actually refers to the Android section of the gradle doc.

This one is fixed.

jwerle:
@mribbons lets review this first thing on Monday (your Tuesday). This looks ready. We just need to test the workflows and discuss how this makes it way into the prebuilds

jwerle:
@mribbons should we close this one too?

PR #78, opened 2022-12-28
Android reworked build

mribbons:
Instabuild!

These changes should be useful for other targets.

android: build cache + no dist delete
Improve Android build time significantly:
> Implement build cache system
  (Gets around ndk rebuild doesn't work, only supports clobber)
  Use SSC_CACHE_ROOT env var to set cache location to somewhere other than
  .config/Socket, otherwise it will be deleted during ./bin/install.sh
> Split libsocket-runtime into 3 libs
  (So changes in one area won't require a complete rebuild)

> Don't delete entire dist folder at build time, keep old build artifacts.
  [android]
  build_remove_path = "dist/android/app/src"

> Add option to disable split runtime libs in android, but still enable lib caching, eg: 
  [android]
  enable_standard_ndk_build = "false"
  disable_split_runtime_libs = "true"

Build time with no changes and preexisting gradle worker is ~5 sec on my mac.

✨  Done in 4.70s.

I'm getting around 15s on windows.

Further notes:
libuv is now libuv.so, APK size is back down to 12MB.

Using  disable_split_runtime_libs = "true" results in a smaller APK than the previous build approach: 9.9mb vs 11mb.

These changes supersede previous PRs:
https://github.com/socketsupply/socket/pull/75
https://github.com/socketsupply/socket/pull/76


jwerle:
@mribbons can we close this?

PR #89, opened 2023-01-05
fix(core/runtime-preload.hh): Android postMessage

### LINKED ISSUE #84: 'window.__ipc.postMessage' does not consider the Android implementation
I see this error when testing an application on Android:

```
VM4:61 Uncaught (in promise) TypeError: Could not determine UserMessageHandler.postMessage in Window
    at Object.value [as postMessage] (<anonymous>:61:13)
    at console.<computed> [as error] (index.js:56115:25)
    at index.js:52122:15
```

This is thrown in an "`else`" situation of this global VM script:

```js
Object.defineProperties(window.__ipc, {                                   
  postMessage: {                                                          
    value: (...args) => {                                                 
      if (window?.webkit?.messageHandlers?.external?.postMessage) {       
        return window.webkit.messageHandlers.external.postMessage(...args)
      }                                                                   
      if (window?.chrome?.webview?.postMessage) {                         
        return window.chrome.webview.postMessage(...args)                 
      }                                                                   
      throw new TypeError(                                                
        'Could not determine UserMessageHandler.postMessage in Window'    
      )                                                                   
    },                                                                    
    enumerable: true                                                      
  }                                                                       
}) 
```

The Android implementation makes `postMessage` available on `window.external`, so this should be the fix:

```js
    if (window?.external?.postMessage) {                         
        return window.external.postMessage(...args)                 
      }
```

jwerle:
- fix missing 'window.external.postMessage()' case for Android

Fixes #84


jwerle:
Looks like it was lost when migrating from io in this commit https://github.com/socketsupply/io/commit/f168cfdeb697909ceae6a9c325c2660a5a1843e8 to socket in this commit https://github.com/socketsupply/socket/commit/f835df9fe3f9e257b86ae743bc12d16d046a67bf

chicoxyzzy (approved):

PR #97, opened 2023-01-10
fix: 'process' leak and 'killProcess' dealloc

jwerle:
This should fix: `libc++abi: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument`


chicoxyzzy (approved):
👍 👍 👍

PR #102, opened 2023-01-13
Custom build source directory with 'SOCKET_HOME'

jwerle:
This pull requests introduces configurable install locations with `SOCKET_HOME` and `PREFIX` (for the `ssc` cli). It also migrates where source files are stored from `~.config/socket` to `~/.local/share/socket` to be more in line with `XDG_DATA_HOME` [1].

This PR also includes a change to support `PREFIX` when installing the `ssc` binary allowing me us to run `./bin/install.sh` **without** `sudo`:

```sh
$ PREFIX=$HOME/.local ./bin/install.sh
```

Which installs `ssc` into my `$HOME/.local/bin` directory, which is in my `PATH`. `PREFIX` may be something other that `/usr/local/` which we default to giving the installer an option to place `ssc` where they'd like.

With the `SOCKET_HOME` environment variable, I can tell the installer where to put the build source files too:

```sh
$ PREFIX=$HOME/.local SOCKET_HOME=$HOME/socket ./bin/install.sh
```

When building an app I can specify where the `SOCKET_HOME` directory is:

```sh
SOCKET_HOME=$HOME/socket ssc build -r -o
```

1. https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -4,9 +4,14 @@ declare root="$(cd "$(dirname "$(dirname "${BASH_SOURCE[0]}")")" && pwd)"
 declare pids=()
 
 LIPO=""
-WORK_DIR=`pwd`
-PREFIX="${PREFIX:-$HOME}"
-BUILD_DIR=$WORK_DIR/build
+declare CWD=`pwd`
+declare PREFIX="${PREFIX:-"/usr/local"}"
+declare BUILD_DIR="$CWD/build"
+declare SOCKET_DIR="${SOCKET_DIR:-"${XDG_DATA_HOME:-"$HOME/.local/share"}/socket"}"
+
+if [ -n "$LOCALAPPDATA" ]; then
+  SOCKET_DIR="$LOCALAPPDATA/Programs/socketsupply"
+fi
 
 if [ ! "$CXX" ]; then
   if command -v clang++ >/dev/null 2>&1; then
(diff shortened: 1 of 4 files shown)


heapwolf (approved):
this is much cleaner, let's ship it!

chicoxyzzy (approved):
it works on my mac

chicoxyzzy:
ignore failing macOS CI, it's fixed

bin/install.sh
```diff
@@ -261,48 +260,59 @@ function _prepare {
...
+
+    cp -f "$BUILD_DIR/$arch-desktop"/bin/* "$SOCKET_HOME/bin"
+    die $? "not ok - unable to move binary into '$SOCKET_HOME'"
+
+    local status="$(ln -sf "$SOCKET_HOME/bin"/* "$PREFIX/bin" 2>&1)"
+    local rc=$?
+
+    if [[ " $status " =~ " Permission denied " ]]; then
+      echo "warn - Failed to link binrary to '$PREFIX/bin': Trying 'sudo'"
+      sudo ln -sf "$SOCKET_HOME/bin"/* "$PREFIX/bin"
```
trevnorris, line 306: Is this necessary? I'd think the script would succeed if they had run it with correct permissions.

trevnorris (approved):
Approved. Have one question, but not a blocker.

bin/install.sh
jwerle: yeah i'll come back to this one!

PR #103, opened 2023-01-13
NPM Modules

### LINKED ISSUE #90: Publish '@socketsupply/socket' on NPM (and GH)
There are two options for releasing on npm:

1. Include all the static libraries for all supported platforms in the npm module.
2. Post-install script to download the artifacts directly from someplace like S3.
3. Publish npm modules for each supported architecture and use a post-install script to `npm i` the module for the specific architecture.

Another consideration is ensuring developers have a comprehensive list of what they need. It's very possible that they npm install ssc just to have building an application fail since they don't have everything necessary.

jwerle:
This pull request fixes #90 and introduces NPM modules based on platform and architecture as optional dependencies to the `@socketsupply/socket` module. Currently, there is only support for `darwin` and `linux`.

## Versioning

The `package.json` for each module is maintained in `npm/packages/@socketsupply/*`. They should be updated manually. However, there is a connivence script for versioning all platform packages:

```sh
bin/version-npm-modules.sh <npm version args>
```

## Publishing

The modules can be published with a helper script `bin/publish-npm-modules.sh [--dry-run] [--only-top-level] [--only-platforms]`

## Installation

Installing the top level module can be done with:

```sh
npm install -g @socketsupply/socket
```

You may need to set your global [NPM Prefix](https://docs.npmjs.com/cli/v8/commands/npm-prefix) first:

**~/.npm-global:**

```sh
npm set prefix $HOME/.npm-global
```

You should include `$HOME/.npm-global/bin` in your `PATH`

**~/.local:**

```sh
npm set prefix $HOME/.local
```

You should include `$HOME/.local/bin` in your `PATH`

Installing with `sudo` should also work - you may need to run it with `sudo -E` if installing on Linux from GitHub Packages

### DIFF
--- a/bin/build-runtime-library.sh
+++ b/bin/build-runtime-library.sh
@@ -16,7 +16,14 @@ declare sources=(
 )
 
 declare arch="$(uname -m)"
+declare host="$(uname -s)"
 declare platform="desktop"
+
+if [[ "$host" = "Linux" ]]; then
+  if [ -n "$WSL_DISTRO_NAME" ] || uname -r | grep 'Microsoft'; then
+    HOST="Win32"
+  fi
+fi
 
 if (( TARGET_OS_IPHONE )); then
   arch="arm64"
@@ -55,7 +62,7 @@ while (( $# > 0 )); do
   args+=("$arg")
 done
 
-if [[ "$(uname -s)" = "Darwin" ]]; then
+if [[ "$host" = "Darwin" ]]; then
   cflags+=("-ObjC++")
   sources+=("$root/src/window/apple.mm")
   if (( TARGET_OS_IPHONE)); then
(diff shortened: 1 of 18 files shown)


chicoxyzzy:
Do we want to keep `ssc` version and package version in sync?

jwerle:
> Do we want to keep `ssc` version and package version in sync?

Yeah that would be ideal! At least the major and minor versions.

cc @trevnorris

chicoxyzzy:
Probably we can remove `VERSION.txt` and `bin/version.sh` and use version from the `package.json`

jwerle:
> Probably we can remove `VERSION.txt` and `bin/version.sh` and use version from the `package.json`

That assumes node or some way to parse the version from `package.json`

heapwolf (approved):

chicoxyzzy (approved):

PR #105, opened 2023-01-15
Process wait

jwerle:
This pull request introduces `Process::wait()` to allow the caller to wait for the child process to exit with a status code. @heapwolf observed a race condition in the build step as the main process (`ssc`) was not blocking after `process->open()` was called.


heapwolf (approved):
FYI im only able to test on macOS atm

PR #117, opened 2023-01-20
Revert "cli: print error when test file doesn't exist"

chicoxyzzy:
Reverts socketsupply/socket#116

This was incorrect. We should check filename from JS at runtime


jwerle (approved):
fair

PR #119, opened 2023-01-24
merge socket api

heapwolf:
this preserves history and merges in socket-api making it easier to sync the javascript api with the runtime.


heapwolf:
CI is probably struggling with the fact that code is in `/api` now

chicoxyzzy (approved):

chicoxyzzy:
docs generation fixes PR #121

api/.eslintrc
```diff
@@ -0,0 +1,9 @@
+{
```
jwerle, line 1: We should move this to the root of the repo

jwerle (approved):

jwerle:
I can follow up with a subsequent PR that can include the `api/` directory in the installation and a copy into the platform specific output path (`socket/` sitting sibling to `index.html|).

Folks should be able to do:

```js
import fs from 'socket/fs.js'
```

PR #120, opened 2023-01-24
Update GH workflow

chicoxyzzy:


heapwolf (approved):
this seems good to me cc @jwerle

jwerle (approved):

PR #122, opened 2023-01-24
Update package.json in the merged repo

chicoxyzzy:


package.json
```diff
@@ -1,29 +1,29 @@
 {
-  "name": "@socketsupply/socket-api",
+  "name": "@socketsupply/socket",
```
jwerle, line 2: We will need to do something with https://github.com/socketsupply/socket/blob/master/npm/packages/%40socketsupply/socket/package.json

jwerle: Actually, since this is private, we can drop the `name` field all together

jwerle (approved):

PR #124, opened 2023-01-27
config: cleanup

### LINKED ISSUE #83: Set default value for 'title' field when missing from 'socket.ini'
Currently, I see `{{title}}` on my Android device for an app I am testing. It isn't defined in `socket.ini` so this makes sense. Ideally, it should fall back to a default value (`name`) or be a **required** field. `name` currently cannot have white spaces, while `title` is a friendly name that can include white space and be seen by humans. Which SHOULD require both IMO, or issue a warning that we're falling back to `name`.

cc @chicoxyzzy @heapwolf

chicoxyzzy:
It's confusing to have all three: `name`, `title` and `executable`

Changes in this PR:
- previously ungrouped fields are in either `[build]` or `[meta]` section now
- new field ~~`app_name`~~ `[debug] name` — string without spaces and special symbols; instead of `name` and `executable`
- new field ~~`app_title`~~ `[meta] title` — any string; instead of `title`
- only ~~`app_name`~~ `[debug] name` is required now; default value for `version` is `1.0.0`
- ~~`app_name`~~ `[debug] name` validation; supports alphanumerical characters, dash and underscore
- ~~`version`~~ `[meta] version` validation; supports semver triple 
- hides `revision` field from the user
- `build` is `[build] script` now

also see https://github.com/socketsupply/socket/pull/124#issuecomment-1408740946

fixes #83

### DIFF
--- a/npm/packages/@socketsupply/socket-darwin-arm64/package.json
+++ b/npm/packages/@socketsupply/socket-darwin-arm64/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@socketsupply/socket-darwin-arm64",
-  "version": "0.0.13",
+  "version": "0.0.14",
   "description": "A Cross-Platform, Native Runtime for Desktop and Mobile Apps — Create apps using HTML, CSS, and JavaScript. Written from the ground up to be small and maintainable.",
   "type": "module",
   "main": "src/index.js",
--- a/npm/packages/@socketsupply/socket-darwin-x64/package.json
+++ b/npm/packages/@socketsupply/socket-darwin-x64/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@socketsupply/socket-darwin-x64",
-  "version": "0.0.4",
+  "version": "0.0.5",
   "description": "A Cross-Platform, Native Runtime for Desktop and Mobile Apps — Create apps using HTML, CSS, and JavaScript. Written from the ground up to be small and maintainable.",
   "type": "module",
   "main": "src/index.js",
--- a/npm/packages/@socketsupply/socket/package.json
+++ b/npm/packages/@socketsupply/socket/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@socketsupply/socket",
-  "version": "0.0.25",
+  "version": "0.0.26",
   "description": "A Cross-Platform, Native Runtime for Desktop and Mobile Apps — Create apps using HTML, CSS, and JavaScript. Written from the ground up to be small and maintainable.",
   "type": "module",
   "main": "index.js",
@@ -38,8 +38,8 @@
   },
   "homepage": "https://github.com/socketsupply/socket#readme",
   "optionalDependencies": {
-    "@socketsupply/socket-darwin-arm64": "^0.0.13",
-    "@socketsupply/socket-darwin-x64": "^0.0.4",
-    "@socketsupply/socket-linux-x64": "^0.0.4"
+    "@socketsupply/socket-darwin-arm64": "^0.0.14",
+    "@socketsupply/socket-darwin-x64": "^0.0.5",
+    "@socketsupply/socket-linux-x64": "^0.0.5"
   }
 }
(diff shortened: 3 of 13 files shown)


chicoxyzzy:
fixes #83

chicoxyzzy:
Window titles should either be set from `<title>` or via JavaScrip, otherwise all windows will start with the name from config which is wrong (there also will be annoying blinking from config name to new name when we need to immediately set it from JS). We removed that option from the config. Title here is a name for the App Store, meta, About window, etc. It's the app title, not window title. `executable` is the same thing as `name`, so I replaced it by `app_name`

If we are going to add `[app]` section, it will look like this:
```ini
[app]
name = my_app_name # in filenames and other places where spaces and special characters are not allowed
title = "My App Name" # in the stores
```

chicoxyzzy:
We already have underscores in other field names
image

At least `bundle_identifier`, `simulator_device`. and `manifest_permissions` have underscore in it

chicoxyzzy:
Let's group remaining fields like this:
```ini
[build]
; An list of environment variables, separated by commas.
env = USER, TMPDIR, PWD
; Set the limit of files that can be opened by your process.
file_limit = 1024
; Advanced Compiler Settings (ie C++ compiler -02, -03, etc).
flags = -O3
; If false, the window will never be displayed.
headless = false
; A directory is where your application's code is located.
input = "src"
; The name of the program and executable to be output. Can't contain spaces or special characters. Required field.
name = "beepboop"
; The binary output path. It's recommended to add this path to .gitignore.
output = "dist"
; The build script
; script = "node build-script.js"

[meta]
; A unique ID that identifies the bundle (used by all app stores).
bundle_identifier = "com.beepboop"
; A string that gets used in the about dialog and package meta info.
copyright = "(c) Beep Boop Corp. 1985"
; A short description of the app.
description = "A UI for the beep boop network"
; Localization
lang = "en-us"
; A String used in the about dialog and meta info.
maintainer = "Beep Boop Corp."
; TODO: maybe the user doesn't need to know about this? 
revision = 123
; The title of the app used in metadata files. This is NOT a window title. Can contain spaces and special characters. Defaults to app_name.
title = "Beep Boop"
; A string that indicates the version of the application. It should be a semver triple like 1.2.3. Defaults to 1.0.0.
version = 1.0.0
```

heapwolf:
I love this, this is really nice

chicoxyzzy:
I think we can get rid of `revision`

heapwolf:
yeah revision is a windows thing, and we dont need to expose it, I think especially since it creates a platform inconsistency. lets just hard code that to `1` and remove it

chicoxyzzy:
We can add versioning to the config, so when people update the version of `ssc` and run it, we can know the version of `config.ini` is not compatible with the version of `ssc` and throw a readable error and maybe even provide an update script.

I think we can add an ungrouped field like this:

```ini
ssc: 1.0.0

[build]
# ...

[meta]
# ...
```

Then,
- if `ssc -v` > config's `ssc` and `ssc` has changed how we work with the config, we print something like `Current version of config is outdated and not compatible with your version of 'ssc'. Please check the release notes (link) or run 'ssc config update'`
- if `ssc -v` < config's `ssc`, we print `Current version of config is newer than installed 'ssc'. Please update your version of 'ssc'.`

heapwolf:
versioning would be :fire:

chicoxyzzy:
I've created a new issue for versioning feature https://github.com/socketsupply/socket/issues/128

jwerle (commented):
I feel like `version` and `copyright` should be properties of the `[app]` section

chicoxyzzy:
Won't adding the `[app]` section be confusing?

Current sections look clear to me https://github.com/socketsupply/socket/pull/124#issuecomment-1408740946

If we'll have `[app]`, I think we need to move `name` there as well.

In the current structure it's clear which fields are used on the build stage and all other fields are used as additional/optional metadata

PR #126, opened 2023-01-29
Native Socket Runtime Javascript API

jwerle:
Native Socket Runtime Javascript API
============================

This pull request introduces a **Native JavaScript API** that is included in the runtime builds. 

## Native JavaScript API

The [runtime api](https://github.com/socketsupply/socket/tree/socket-runtime-api-build/api) is installed into `SOCKET_HOME` as `api/` and included in the build. The various APIs can be imported in ESM with `import module from 'socket:module'` syntax:

```js
import crypto from 'socket:crypto'
import fs from 'socket:fs'

// ...
```

This works on Android, Linux, iOS, and macOS. A custom `socket:` protocol is used to provide the ESM modules. The specifies for getting it to work is outlined below:

**Android** 
An [importmap](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) is injected into the page _before any other javascript is included_ to make this work as overloading `socket:` requests was not sufficient as a "base URL" could not be determined, causing ESM modules to fail to load. We [generate](https://github.com/socketsupply/socket/blob/socket-runtime-api-build/bin/generate-api-import-map.sh) an `importmap.json` and place it in `SOCKET_HOME/api` to be read and injected inline into the user's HTML (see `importmap.json` below).

**Linux**
A custom [Webkit2GTK](https://webkitgtk.org/reference/webkit2gtk/stable/index.html) [native extension](https://github.com/socketsupply/socket/blob/socket-runtime-api-build/src/extensions/linux/webkit.cc) was created so that it could call `webkit_uri_request_set_uri()` in the actual webkit webview process, prior to the native side interception. This allows `socket:` URLs to look like `file://` URLs to the webview and native runtime, preserving the origin and base URL

**iOS/macOS**
A [custom scheme handler](https://github.com/socketsupply/socket/blob/socket-runtime-api-build/src/ipc/bridge.cc#L1243-L1304) for the `socket:` protocol was introduced with specific care for resolving native ESM modules and preventing duplicate requests of the same module: `socket:crypto` and `./crypto.js` (relative). An inline proxy module is created and returned in the responses for ESM modules so imports of the module of concern are imported exactly once at the canonical URL (file:///...).

The template for the proxy ESM module looks like:

```js
export * from '{{url}}'
const exports = await import('{{url}}');
export default exports.default ?? undefined
```

**Windows**
An implementation for windows is still missing, but this change does not break the runtime or prevent usage of the javascript API. Windows users will **not** be able to `import fs from 'socket:fs'` (yet) until we can implement a custom `socket:` protocol. I also suspect we'll need to support an `importmap` on Windows. We may need to [inject it like on Android](https://github.com/socketsupply/socket/blob/socket-runtime-api-build/src/android/window.kt#L52-L70) 

### `importmap.json`

```json
{
  "imports": {
     "socket:backend": "./socket/backend.js",
     "socket:bluetooth": "./socket/bluetooth.js",
     "socket:bootstrap": "./socket/bootstrap.js",
     "socket:buffer": "./socket/buffer.js",
     "socket:console": "./socket/console.js",
     "socket:crypto": "./socket/crypto.js",
     "socket:dgram": "./socket/dgram.js",
     "socket:dns": "./socket/dns.js",
     "socket:errors": "./socket/errors.js",
     "socket:events": "./socket/events.js",
     "socket:fs": "./socket/fs.js",
     "socket:gc": "./socket/gc.js",
     "socket:index": "./socket/index.js",
     "socket:ipc": "./socket/ipc.js",
     "socket:net": "./socket/net.js",
     "socket:node-esm-loader": "./socket/node-esm-loader.js",
     "socket:os": "./socket/os.js",
     "socket:p2p": "./socket/p2p.js",
     "socket:path": "./socket/path.js",
     "socket:polyfills": "./socket/polyfills.js",
     "socket:process": "./socket/process.js",
     "socket:runtime": "./socket/runtime.js",
     "socket:stream": "./socket/stream.js",
     "socket:util": "./socket/util.js",
     "socket:crypto/sodium": "./socket/crypto/sodium.js",
     "socket:dns/index": "./socket/dns/index.js",
     "socket:dns/promises": "./socket/dns/promises.js",
     "socket:fs/binding": "./socket/fs/binding.js",
     "socket:fs/constants": "./socket/fs/constants.js",
     "socket:fs/dir": "./socket/fs/dir.js",
     "socket:fs/fds": "./socket/fs/fds.js",
     "socket:fs/flags": "./socket/fs/flags.js",
     "socket:fs/handle": "./socket/fs/handle.js",
     "socket:fs/index": "./socket/fs/index.js",
     "socket:fs/promises": "./socket/fs/promises.js",
     "socket:fs/stats": "./socket/fs/stats.js",
     "socket:fs/stream": "./socket/fs/stream.js",
     "socket:path/index": "./socket/path/index.js",
     "socket:path/path": "./socket/path/path.js",
     "socket:path/posix": "./socket/path/posix.js",
     "socket:path/win32": "./socket/path/win32.js",
     "socket:test/tap": "./socket/test/tap.js",
     "socket:external/libsodium/index": "./socket/external/libsodium/index.js",
     "socket:test/tap/fast-deep-equal": "./socket/test/tap/fast-deep-equal.js",
     "socket:test/tap/harness": "./socket/test/tap/harness.js",
     "socket:test/tap/index": "./socket/test/tap/index.js"
  }
}
```

### `@socketsupply/socket`

Runtime modules can also be imported from NPM. This replaces the `@socketsupply/socket-api` module

```js
import fs from '@socketsupply/socket/fs.js'
```

## `sodium` and `tapzero` builtins

[libsodium.js](https://github.com/jedisct1/libsodium.js/) and [tapzero](https://github.com/socketsupply/tapzero) are built in and can be imported:

```js
import sodium from 'socket:crypto/sodium'
```

```js
import { test } from 'socket:test/tap'
```


chicoxyzzy:
linter fixes https://github.com/socketsupply/socket/pull/127

chicoxyzzy:
Why is `stream-relay` an external dependancy? Is the plan to move it to the `socket` repo?

heapwolf:
> Why is `stream-relay` an external dependancy? Is the plan to move it to the `socket` repo?

its a submodule because it's 1. experimental-stage, 2. should also run on nodejs and deno and we don't want to maintain two copies of it.

chicoxyzzy:
What if we'll prefix our node-compatible internal modules with `node:`? This can make the code more portable between Socket and other platforms. I'm trying to find if Bun supports `node:` prefix, but can't find it in the docs

test/src/test-context.js
```diff
@@ -1,10 +1,10 @@
...
-import process from '../../api/process.js'
-import '../../api/runtime.js'
+import { GLOBAL_TEST_RUNNER } from 'socket:test/tap'
+import console from 'socket:console'
+import process from 'socket:process'
+import 'socket:runtime'
 
 // uncomment below to get IPC debug output in stdout
-// import ipc from '../../ipc.js'
+// import ipc from '../../api/ipc.js'
```
chicoxyzzy, line 7:
```suggestion
// import ipc from '.socket:ipc'
```

chicoxyzzy (commented):
There are a few mentions of `socket-api` left:
- `README.md`
- `api/dns.js`
- `api/fs.js`
- `api/README.md`
- `api/fs/index.js`
- `test/socket.ini`
- `test/src/frontend/index.html`

README.md
```diff
@@ -22,4 +22,4 @@ See the [`Socket API`][0] library for the unit and integration test suite.
 If you want to contribute to the Socket Runtime project itself, please connect
 with any current project contributor.
 
-[0]:https://github.com/socketsupply/socket-api
+[0]:https://github.com/socketsupply/socket
```
chicoxyzzy, line 25: Should we remove the whole `Testing` section? It mentions `Socket API` as a library

chicoxyzzy (approved):

Issue #113, opened 2023-01-17
Document exported globals


trevnorris:
Currently, socket exports `window.__args` and `window.__ipc`. If these are going to be considered public then we should document them.

Ref: https://github.com/socketsupply/socket/issues/14

chicoxyzzy:
We try to move these two globals to the "native" Socket modules https://github.com/socketsupply/socket/pull/126

chicoxyzzy:
`window.__ipc` doesn't depend on any arguments, so it could go to the `socket:ipc/runtime`

Most of `window.__args` values could be written to some module with exported constants. It seems that only `window.__args.index` can't be shared between all the windows, because it's a window id

trevnorris:
Excellent. Looks like this means there won't be any global exports. I'll close this issue as resolved.

PR #130, opened 2023-01-31
preload script cleanup (part 1)

chicoxyzzy:
From https://github.com/socketsupply/socket/issues/113#issuecomment-1409857594

> `window.__ipc.postMessage` code doesn't depend on any Socket arguments, so it could go to the `socket:ipc`'s exported `postMessage` function

- [x] done

> Most of `window.__args` values could be written to some module with exported constants. It seems that only `window.__args.index` can't be shared between all the windows, because it's a window id

- [x] `__args.os`, `__args.cwd`
- [x] `__args.version`
~~- [ ] `__args` that use `opts`: `argv`, `config`, `debug`, `env`~~ (will move them to part 2)
~~- [ ] `__args.index`~~ (will move it to part 2)
~~- [ ] `DOMContentLoaded` event handler~~ (will move it to part 2)


chicoxyzzy:
@jwerle I want to make these changes before docs update and I need your help here. How can we pass the `__args` values from C++ to JavaScript?
Also I think this code JS code will never work, cause we don't use `window.parent` anymore. What is it supposed to do?

chicoxyzzy:
It's also confusing that we have some of [these options](https://github.com/socketsupply/socket/blob/master/src/window/options.hh#L7-L30) on `Window`.

```c++
    int debug = 0;
    int port = 0;
    bool isTest = false;
    bool headless = false;
    String cwd = "";
    String argv = "";
    String preload = "";
    String env;
    Map appData;
```
This options are related to the app more than window

chicoxyzzy:
> How can we pass the __args values from C++ to JavaScript?

I think we separate C++ and JS by adding corresponding IPC messages to get values and cache them in JS on first call for faster access

jwerle:
> I'm going to use the same approach as in [`api/fs/constants.js`](https://github.com/socketsupply/socket/blob/master/api/fs/constants.js#L3)

👍 Yep that should work!

chicoxyzzy:
For some reason after https://github.com/socketsupply/socket/pull/130/commits/17574c3b7201354222757359d1fe39f8cce3eef4 test app exits before the FS tests

chicoxyzzy:
I just rebased on master and now FS tests are last but one. Test runner exits before FS tests again. I think that the timeout is set to too small value here https://github.com/socketsupply/socket/commit/36cd3b3b0428afcfcbc7beb420336306f7656a9b

chicoxyzzy:
test runner fix https://github.com/socketsupply/socket/pull/143

chicoxyzzy:
I'm going to move the rest of preload stuff out of global variables in the next PR as this one contains too many changes, fixes and new tests already. Also these changes are enough to work on a new the public API

chicoxyzzy:
Is it safe to remove this code? https://github.com/socketsupply/socket/blob/master/src/android/bridge.kt#L194-L201

Except that, I think PR is ready for merge

chicoxyzzy:
On master branch Windows tests are passing in 5-6 minutes. On this branch CI exits after timeout (12 min)

heapwolf (approved):
I like this approach, overall seems good to me. let's ship it.

src/ipc/bridge.cc
```diff
@@ -835,26 +818,27 @@ void initFunctionsTable (Router *router) {
...
-        {"data", cwd}
-      };
-    }
-
+  router->map("platform.primordials", [=](auto message, auto router, auto reply) {
+    std::regex platform_pattern("^mac$", std::regex_constants::icase);
+    auto platformRes = std::regex_replace(platform.os, platform_pattern, "darwin");
+    auto arch = std::regex_replace(platform.arch, std::regex("x86_64"), "x64");
+    arch = std::regex_replace(arch, std::regex("x86"), "ia32");
+    arch = std::regex_replace(arch, std::regex("arm(?!64).*"), "arm64");
```
jwerle, line 828: Should be `"arm"`, not `"arm64"` as the replacement

PR #133, opened 2023-02-01
Adds a symmetrical API for user-space IPC

heapwolf:
On the node.js or deno side...

```js
import { IPC } from '@socketsupply/socket/ipc.js'
const ipc = new IPC()

const value = { foo: true }

// send to window 0, expect a response
const result = await ipc.request({ window: 0, value })

// send to all windows, (assume window 0) expect a response
const res = await ipc.request({ value })

// send to window 1, don't expect a response
const ipc.send({ window: 1, value })

// send to window 0, don't expect a response
const ipc.send({ value })

// a request to respond to an ipc message ("req" includes the index and seq)
ipc.on('request', (req, value) => {
  ipc.respond(req, value)
})
```

on the Socket side...

```js
import { IPC } from 'socket:ipc'
const ipc = new IPC()

const value = { foo: true }

// send to window 0, expect a response
const res = await ipc.request({ value })

// send to the backend
const ipc.send({ value })

// send to window 2
const ipc.send({ window: 2, value })

// a request to respond to an ipc message ("req" includes the index and seq)
ipc.on('request', (req, value) => {
  ipc.respond(req, value)
})
```

### DIFF
--- a/api/ipc.js
+++ b/api/ipc.js
@@ -1,27 +1,21 @@
 /**
  * @module IPC
  *
- * This is a low level API which you don't need unless you are implementing
- * a library on top of Socket SDK. A Socket SDK app has two or three processes.
+ * The `Frontend` process is the UI where the HTML, CSS and JS runs. In
+ * Socket, there is native File IO, UDP, and Bluetooth, so there is less
+ * need to run a backend.
  *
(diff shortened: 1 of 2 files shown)


chicoxyzzy:
covered by https://github.com/socketsupply/socket/pull/210

PR #137, opened 2023-02-02
Move ssc-node to the @socketsupply/socket

chicoxyzzy:
## Allow CommonJS

Experiment with npm workspaces and module exports. This allows to symlink the `@socketsupply/socket` from the test app. I moved `ssc-node` to the `node` directory of `@socketsupply/socket`. I use `ssc-node` and not https://github.com/socketsupply/socket/pull/133 to be sure all tests are passing. This patch in `@socketsupply/socket`'s `package.json` allows us to enable usage of both CommonJS and ESM

```json
  "exports": {
    "import": "./node/index.js",
    "require": "./node/index.cjs"
  },
```

CommonJS **is not possible** in the https://github.com/socketsupply/socket/pull/133 draft.

This approach also allows us to split the frontend IPC and backend IPC and reuse the same modules. If necessary they can share the base class, but we will need to move the `/api` files to the `/npm/packages/@socketsupply/socket` (what in fact looks like a proper thing to do to me anyway! cc @jwerle)

# Disadvantages of moving Node.js code to the `api`

1. One **big** disadvantage is that `@socketsupply/socket`+ it's platform-specific dependancy are a very heavyweight packages and they will be added to **each** Socket Runtime project even if Socket Runtime already installed globally. For `darwin-x64` it's almost 18 MB and for `darwin-arm64` it's almost 28 MB!

1. Another disadvantage is that imports are different for frontend usage and backend usage.
```js
// frontend
import { ipc } from 'socket:ipc'
```
```js
// backend
import { ipc } from '@socketsupply/ipc'
```

That looks like **two different dependancies**. Import maps are not supported in Node.js yet, so we can't make it `socket:ipc` there ([Deno has](https://deno.land/manual@v1.30.0/basics/import_maps) import maps support). 

# Summary

Given these two disadvantages, I'd prefer to have a **separate tiny module** named `@socketsupply/socket-ipc-node` (ex-[`@socketsupply/ssc-node`](https://github.com/socketsupply/ssc-node)) for Node.js. This module also should mirror the `socket:ipc` API and have both ESM and CommonJS support.

### DIFF
--- a/npm/packages/@socketsupply/socket/node/index.js
+++ b/npm/packages/@socketsupply/socket/node/index.js
@@ -1,0 +1,306 @@
+// @ts-check
+import { format } from 'node:util'
+import { writeSync, fsyncSync } from 'node:fs'
+
+const AUTO_CLOSE = process.env.AUTO_CLOSE
+const MAX_MESSAGE_KB = 512 * 1024
+
+function isObject (o) {
+  return o && typeof o === 'object' && !Array.isArray(o)
+}
+
+const write = s => {
+  if (s.includes('\n')) {
+    throw new Error('invalid write()')
+  }
+
+  if (s.length > MAX_MESSAGE_KB) {
+    const len = Math.ceil(s.length / 1024)
+    process.stderr.write('WARNING: Sending large message to webview: ' + len + 'kb\n')
+    process.stderr.write('RAW MESSAGE: ' + s.slice(0, 512) + '...\n')
+  }
+
+  return new Promise(resolve =>
+    process.stdout.write(s + '\n', resolve)
+  )
+}
+
+console.log = (...args) => {
+  const s = args.map(v => format(v)).join(' ')
+  const enc = encodeURIComponent(s)
+  // fs.appendFileSync('tmp.log', s + '\n')
+  write(`ipc://stdout?value=${enc}`)
+}
+console.error = console.log
+
+process.on('exit', (exitCode) => {
+  const seq = String(ipc.nextSeq++)
+
+  let value = new URLSearchParams({
+    index: '0',
+    seq,
+    value: String(exitCode)
+  }).toString()
+
+  value = value.replace(/\+/g, '%20')
+
+  writeSync(1, `ipc://exit?${value}\n`)
+  try {
+    fsyncSync(1)
+  } catch (_) {
+    // fsync(1) can fail in github actions for reasons unclear.
+    // maybe the stdout is weird in that environment.
+  }
+})
+
+//
+// Internal IPC API
+//
+const ipc = { nextSeq: 0 }
+
+ipc.resolve = async (seq, state, value) => {
+  const method = !Number(state) ? 'resolve' : 'reject'
+  if (!ipc[seq] || !ipc[seq][method]) return
+
+  /**
+   * TODO Who handles this error ?
+   */
+  try {
+    await ipc[seq][method](value)
+  } finally {
+    delete ipc[seq]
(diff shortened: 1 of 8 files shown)


npm/packages/@socketsupply/socket/node/index.js
```diff
@@ -0,0 +1,306 @@
...
+    process.stderr.write('RAW MESSAGE: ' + s.slice(0, 512) + '...\n')
+  }
+
+  return new Promise(resolve =>
+    process.stdout.write(s + '\n', resolve)
+  )
+}
+
+console.log = (...args) => {
+  const s = args.map(v => format(v)).join(' ')
```
jwerle, line 29: Format string in first argument is dropped.

Try

```js
util.format(...args)
```

npm/packages/@socketsupply/socket/node/index.js
```diff
@@ -0,0 +1,306 @@
...
+  const enc = encodeURIComponent(s)
+  // fs.appendFileSync('tmp.log', s + '\n')
+  write(`ipc://stdout?value=${enc}`)
+}
+console.error = console.log
```
jwerle, line 34: There is a `ipc://stderr` route

npm/packages/@socketsupply/socket/node/index.js
```diff
@@ -0,0 +1,306 @@
...
+// Internal IPC API
+//
+const ipc = { nextSeq: 0 }
+
+ipc.resolve = async (seq, state, value) => {
+  const method = !Number(state) ? 'resolve' : 'reject'
+  if (!ipc[seq] || !ipc[seq][method]) return
+
+  /**
+   * TODO Who handles this error ?
```
jwerle, line 66: Probably `ipc` if it was an event emitter!

npm/packages/@socketsupply/socket/node/index.js
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  await write(`ipc://${cmd}?${value}`)
+  return promise
+}
+
+ipc.send = async o => {
+  try {
+    // TODO: use structuredClone instead once we are on node 17+
+    o = JSON.parse(JSON.stringify(o))
```
jwerle, line 111: You'd be surprised to know that `structuredClone` is much slower [1]

1. https://www.measurethat.net/Benchmarks/Show/18541/0/jsonstringify-vs-structuredclone#latest_results_block

chicoxyzzy:
@jwerle code in the `node` repo is just a copy of `ssc-node`. I don't want to move it to the `socket` repo. Here I did it in order to try how npm workspaces work. I don't want to merge this branch. Although, I think that we should move `/api` files to the `/npm/packages/@socketsupply/socket` repo. Please read the issue description.

jwerle:
> @jwerle code in the `node` repo is just a copy of `ssc-node`. I don't want to move it to the `socket` repo. Here I did it in order to try how npm workspaces work. I don't want to merge this branch. Although, I think that we should move `/api` files to the `/npm/packages/@socketsupply/socket` directory. Please read the issue description.

yes, sorry. I clearly went into auto review mode without finishing

I don't think we should move the `api/` source files into the `npm/` directory. The NPM module should be considered a superset of the runtime js api.

chicoxyzzy:
Ok, I think we have consensus: we want a new module that will mirror the API from #133. We can put this module in the monorepo to make it easier to introduce changes along with `socket:ipc` and test app updates 

I'm going to close this PR

PR #143, opened 2023-02-04
api/test: fix test runner

chicoxyzzy:
All the tests should run on the same tick, otherwise they don't run at all

Better review with hidden spaces: https://github.com/socketsupply/socket/pull/143/files?diff=split&w=1

### DIFF
--- a/api/test/index.js
+++ b/api/test/index.js
@@ -4,9 +4,7 @@ import deepEqual from './fast-deep-equal.js'
 import process from '../process.js'
 import os from '../os.js'
 
-const {
-  SOCKET_TEST_RUNNER_TIMEOUT = getDefaultTestRunnerTimeout()
-} = process.env
+const SOCKET_TEST_RUNNER_TIMEOUT = getDefaultTestRunnerTimeout()
 
 const NEW_LINE_REGEX = /\n/g
 const OBJ_TO_STRING = Object.prototype.toString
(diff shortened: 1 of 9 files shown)


test/src/index.js
```diff
@@ -13,9 +12,5 @@ import './dns.js'
 import './backend.js'
 import './crypto.js'
 import './util.js'
-
-// TODO(@jwerle): FIXME
-if (os.platform() !== 'win32') {
-  await import('./runtime.js')
-  await import('./fs.js')
-}
```
chicoxyzzy, line 21: these tests should run in the same tick

api/test/index.js
```diff
@@ -188,7 +188,6 @@ export class Test {
     try {
       fn()
     } catch (err) {
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
chicoxyzzy, line 191: otherwise standard yells on me

chicoxyzzy:
Windows Ci never fail on tests because we can't get the exit code 1 from the test app. I will remove win check from the PR.

test/src/index.js
```diff
@@ -13,9 +12,5 @@ import './dns.js'
...
+import './runtime.js'
```
jwerle, line 15: we had this check here for windows, why is it removed? imports are not synchronous so with

chicoxyzzy: I moved these checks to the modules. Tests may end with `exit 0` early without this like [here](https://github.com/socketsupply/socket/actions/runs/4099615181/jobs/7069821754). There are failing tests, but no exit 1 as runner exits on some of previous ticks

chicoxyzzy: so this gives a false positive

Issue #150, opened 2023-02-06
install.sh compiler test doesn't work on windows


mribbons:
```
clang++ -x c++ - -Wno-unused-command-line-argument -stdlib=libstdc++ -std=c++2a -I/d/code/socket/socket/include -I/d/code/socket/socket/build/uv/include -I/d/code/socket/socket/build/include -I/d/code/socket/socket/ 
src -DSSC_BUILD_TIME=1675680133 -DSSC_VERSION_HASH=62576ab -DSSC_VERSION=0.1.0 -D_MT -D_DLL -DWIN32 -DWIN32_LEAN_AND_MEAN -Xlinker /NODEFAULTLIB:libcmt -Wno-nonportable-include-path -g -O0 < build/main.cc
--9b07f4.o : error LNK2019: unresolved external symbol atexit referenced in function "void __cdecl SSC::`dynamic initializer for 'platform''(void)" (??__Eplatform@SSC@@YAXXZ)
--9b07f4.o : error LNK2019: unresolved external symbol memcpy referenced in function "protected: void __cdecl std::time_get<char,class std::istreambuf_iterator<char,struct std::char_traits<char> > >::_Getvals<wchar_t>(wchar_t,class std::_Locinfo const &)" (??$_Getvals@_W@?$time_get@DV?$istreambuf_iterator@DU?$char_traits@D@std@@@std@@@std@@IEAAX_WAEBV_Locinfo@1@@Z)
--9b07f4.o : error LNK2019: unresolved external symbol __imp_wcslen referenced in function "wchar_t * __cdecl std::_Maklocwcs(wchar_t const *)" (?_Maklocwcs@std@@YAPEA_WPEB_W@Z)
--9b07f4.o : error LNK2019: unresolved external symbol __imp_calloc referenced in function "wchar_t * __cdecl std::_Maklocwcs(wchar_t const *)" (?_Maklocwcs@std@@YAPEA_WPEB_W@Z)
--9b07f4.o : error LNK2019: unresolved external symbol "void __cdecl operator delete(void *,unsigned __int64)" (??3@YAXPEAX_K@Z) referenced in function "void __cdecl std::_Deallocate<16,0>(void *,unsigned __int64)" (??$_Deallocate@$0BA@$0A@@std@@YAXPEAX_K@Z)
--9b07f4.o : error LNK2001: unresolved external symbol __CxxFrameHandler3
[...]

PR #92, opened 2023-01-09
install.ps1: resolve issues

mribbons:
Get-Command wasn't detecting absense of commands
libuv requires nmake on windows, install vs_build tools to get it Don't use LLVM/Clang

PR #151, opened 2023-02-06
Windows install sh

mribbons:
Windows: use install.sh
install.sh now called by install.ps1

Usage:
```
.\bin\install.ps1
```

Notes:

``mingw sh`` required - Currently install.ps1 just looks for %ProgramFiles\Git\bin\sh.exe
``skipwebview`` - Not Required, WebView2LoaderStatic.lib existence is checked, downloaded by ``install.sh`` (windows only)
``debug`` - Now governed by DEBUG enviroment variable from within ``install.sh``, ``cflags.sh``

We still need to merge this one:
https://github.com/socketsupply/socket/pull/92

Compiler check not working on windows:
This check is really slow anyway, can we come up with a different test file than ``src\common.hh``?
https://github.com/socketsupply/socket/issues/150


heapwolf (changes requested):
id say if we can get CI passing this is good to go

PR #160, opened 2023-02-08
Fix Windows tests

trevnorris:
I'm not proud of how backslashes are escaped. All tests, except the runtime tests, are now passing.

I also believe there's a memory leak somewhere. The test runner requires > 900MB memory to execute. That doesn't sound right.

### DIFF
--- a/test/src/ipc.js
+++ b/test/src/ipc.js
@@ -82,10 +82,17 @@ if (window.__args.os !== 'ios' && window.__args.os !== 'android') {
     const response = ipc.sendSync('test', { foo: 'bar' })
     t.ok(response instanceof ipc.Result)
     const { err } = response
-    t.equal(err?.toString(), 'NotFoundError: Not found')
+    // Make lower case to adjust for implementation differences.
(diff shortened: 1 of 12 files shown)


test/src/ipc.js
```diff
@@ -82,10 +82,17 @@ if (window.__args.os !== 'ios' && window.__args.os !== 'android') {
...
+    // Make lower case to adjust for implementation differences.
+    t.equal(err?.toString().toLowerCase(), 'notfounderror: not found')
     t.equal(err?.name, 'NotFoundError')
-    t.equal(err?.message, 'Not found')
-    t.ok(err?.url.startsWith('ipc://test?foo=bar&index=0&seq=R'))
+    // Make lower case to adjust for implementation differences.
+    t.equal(err?.message.toLowerCase(), 'not found')
+    // win32 adds on the trailing slash in the URL.
+    if (window.__args.os === 'win32') {
+      t.ok(err?.url.startsWith('ipc://test/?foo=bar&index=0&seq=R'))
```
jwerle, line 92: we'll eventually fix and this will fail 🗡️

jwerle (approved):

PR #161, opened 2023-02-09
Static ESM

jwerle:
This pull request ensures all ESM in the JavaScript API and IPC Bridge ESM proxy are static and can be loaded before `window.onload` is called. This means no dynamic imports (`await import()`) which will block


src/ipc/bridge.cc
```diff
@@ -1309,9 +1309,9 @@ export default exports.default ?? undefined
     // URLs (socket:...)
     auto moduleTemplate =
 R"S(
+import module from '{{url}}'
 export * from '{{url}}'
-const exports = await import('{{url}}');
-export default exports.default ?? undefined
+export default module
```
chicoxyzzy, line 1314: we can save the template to a variable to reuse in both line 1223 and 1310

chicoxyzzy (approved):
👍

heapwolf (approved):

src/ipc/bridge.cc
jwerle: done

PR #162, opened 2023-02-10
chore: remove "socket:index"

chicoxyzzy:
In the past we have used `index.js` to export everything and add `window.socket` global variable. This is not what we are doing now. Also this code throws.
```js
import socket from 'socket:index'
```

It's safe to remove the file


jwerle (approved):

heapwolf (commented):
🏁

PR #166, opened 2023-02-13
Windows Prebuilds / Create Socket App works on Win32

mribbons:
Includes:
1. Windows prebuilds: 
   ``sh bin\publish-npm-modules.sh``
2. @socketsupply/socket-win32-x64 now installs pre reqs for `ssc build` on windows
3. install.ps1 no longer uses `chocolatey`

This PR incorporates:
https://github.com/socketsupply/socket/pull/92
https://github.com/socketsupply/socket/pull/151


npm/packages/@socketsupply/socket-win32-x64/package.json
```diff
@@ -12,7 +12,11 @@
   ],
   "scripts": {
     "test": ":",
-    "install": "cd bin && node verify-platform.js win32 x64"
+    "preinstall": "node ./bin/install-pre-reqs.js && echo Pre requisite checks complete.\nUse\n> install-pre-reqs\nto run this again.",
+    "install": "cd bin && node verify-platform.js win32 x64 -"
+  },
+  "bin": {
```
jwerle, line 18: No need to add it to `bin`

jwerle (approved):
Do we need the VS files in `bin/`?

npm/packages/@socketsupply/socket-win32-x64/package.json
mribbons: Hey @jwerle 
My thought is that ap devs can run
```sh
npx @socketsupply/socket-win32-x64
```

If they want to rerun the dependency installer.

mribbons:
> Do we need the VS files in `bin/`?

We could put them somewhere else, but they are required by install.ps1.

We could package `.vsconfig-csa` only.

npm/packages/@socketsupply/socket-win32-x64/package.json
jwerle: ah! I like this! Maybe we can just call the bin `socket-win32-x64-setup` and do the same thing for the others?

src/cli/cli.cc
```diff
@@ -1771,23 +1763,15 @@ int main (const int argc, const char* argv[]) {
...
-            log("build failed, exiting with code " + code);
-            // TODO(trevnorris): Force non-windows to exit the process.
-            exit(std::stoi(code));
-          }
-        }
-      );
-
-      process->open();
-      process->wait();
+      // Removed new SSC::Process() style code, caused random crashes
```
jwerle, line 1766: cc @heapwolf @chicoxyzzy @trevnorris

jwerle:
> > Do we need the VS files in `bin/`?
> 
> We could put them somewhere else, but they are required by install.ps1.
> 
> We could package `.vsconfig-csa` only.

oh if they are required, then we can keep them there. Should we move them to the root of the repo, or do they need to sit sibling to the ps1 files?

src/cli/cli.cc
chicoxyzzy: IIRC the process wrapper was necessary to get the stdout and stderr

mribbons:
> or do they need to sit sibling to the ps1 files?

They don't need to be siblings, could move them to root, it's means more clutter though. They are install related so I thought it's better to keep them in bin in lack of any better option.

src/cli/cli.cc
mribbons: @chicoxyzzy stderr is being redirected to stdout now:
https://github.com/socketsupply/socket/blob/create-socket-win32-setup/src/process/process.hh#L51
```
inline ExecOutput exec (SSC::String command) {
    command = command + " 2>&1";
```

stdout goes to ExecOutput.output and will be written to ssc stdout if error code != 0.

heapwolf: ExecOutput provides streamed (not blocking) output, and exit code handling consistently across all platforms, exec will block and not show output to the user so they won't know if their script is making progress.

heapwolf (changes requested):

src/cli/cli.cc
heapwolf: we should just figure out the root problem, what caused the random crashes, because removing this code will break a lot of other use cases.

trevnorris (commented):
Awesome stuff. Still testing out the branch. Wanted to post my first review while doing that.

npm/packages/@socketsupply/socket-win32-x64/package.json
mribbons: @jwerle What would the npx call and bin look like exactly?

bin/install.sh
```diff
@@ -117,34 +145,77 @@ function _build_cli {
...
+    echo "$BUILD_DIR/lib/$arch-$platform/WebView2LoaderStatic.lib exists."
+    return
+  fi
+
+  local tmp=$(mktemp -d)
+  local pwd=$(pwd)
+
+  echo "# Downloading Webview2"
+
+  curl -L https://www.nuget.org/api/v2/package/Microsoft.Web.WebView2/1.0.1619-prerelease --output $tmp/webview2.zip
```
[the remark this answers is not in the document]
mribbons: I thought about this, but recall that the goal here is to have one single code path.
Any duplicated code in `install.ps1` can be removed. 
I left the ps1 build process there for reference, but due to the fact that it can't work without implemented more duplicated code (Copying the `.a` etc), this code is basically vestigial at this point.

Note that we also have the same problem with `libuv` URI, which was actually coming from different sources in `install.ps1` and `install.sh`.

I just removed all unrequired code from `install.ps1` that I'm aware of.

npm/src/index.js
```diff
@@ -1,7 +1,7 @@
 import path from 'node:path'
 import os from 'node:os'
 
-const dirname = path.dirname(import.meta.url.replace('file://', ''))
+const dirname = path.dirname(import.meta.url).replace(`file://${os.platform() === 'win32' ? '/' : ''}`, '')
```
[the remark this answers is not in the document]
mribbons: Marked for future work: https://github.com/socketsupply/socket/issues/183

chicoxyzzy (commented):
That's awesome!

Can we make Windows CI red on failing tests?

mribbons:
> Can we make Windows CI red on failing tests?

It's an issue for a separate branch, but I will take a look now.

npm/packages/@socketsupply/socket-win32-x64/package.json
jwerle:
```json
"bin": {
  "socket-win32-x64": "bin/socket-win32-x64-setup"
}
```

```sh
npx @socketsupply/socket-win32-x64 # executes the `socket-win32-x64` bin
```

mribbons:
@chicoxyzzy 
> > Can we make Windows CI red on failing tests?
> 
> It's an issue for a separate branch, but I will take a look now.

https://github.com/socketsupply/socket/pull/184 !!

chicoxyzzy (approved):

trevnorris (approved):
I haven't tested from a clean machine to see if it sets everything up correctly, but everything else appears to work.

PR #170, opened 2023-02-15
ssc build: fix race condition crash

mribbons:
Process handle is created in Process::open()
Process status is then checked in a thread
By the time this check happens, Process::open() has completed, resulting in an invalid handle GetExitCodeProcess() would then return a spurious value of 4294967295, or sometimes other values. stoi() would then crash with an out of range exception.

Solution:
1. Pass copy of thread handle to process status waiter thread
2. On GetExitCodeProcess() failure, call GetLastError() and write assosciated error message to stderr The output will appear as follows:
```
• package prepared 683ms
Error 6 in SSC::Process::open() GetExitCodeProcess(): The handle is invalid.

• ran user build command 19ms
• build failed, exiting with code 4294967295 0ms
```

3. Use std::stoul() instead of std::stoi() in case any large return codes occur in other scenarios.

### DIFF
--- a/src/desktop/main.cc
+++ b/src/desktop/main.cc
@@ -142,10 +142,12 @@ MAIN {
     // launched from the `ssc` cli
     app.fromSSC = s.find("--from-ssc") == 0 ? true : false;
 
+    #ifdef _WIN32
     if (!app.w32ShowConsole && s.find("--w32-console") == 0) {
       app.w32ShowConsole = true;
       app.ShowConsole();
     }
+    #endif
 
     if (s.find("--test") == 0) {
       suffix = "-test";


src/process/win.cc
```diff
@@ -7,6 +7,28 @@
 
 namespace SSC {
 
+SSC::String FormatError(DWORD error, SSC::String source)
```
jwerle, line 10: can you format this to:

```c++
SSC::String FormatError (DWORD error, SSC::String source) {
```

jwerle (approved):

chicoxyzzy:
Linux build is failing

jwerle:
> Linux build is failing

unrelated!

jwerle (approved):

src/process/win.cc
mribbons: Done

PR #171, opened 2023-02-15
Desktop Window refactor

chicoxyzzy:
It's not ready for review yet. I want to test things on CI (at some point) and let people see what upcoming changes will be (see `test/application.js` and `test/window/js`)

Tests could (and probably will) now as I run `ssc build` for now and don't care about the standard linter most of the time so I can see the Web Inspector logs


api/application.js
```diff
@@ -0,0 +1,289 @@
...
+  const windows = await getWindows([index])
+  return windows[index]
+}
+
+/**
+ * Returns the ApplicationWindow instance for the current window.
+ * @return {Promise<ApplicationWindow>}
+ */
+export async function getCurrentWindow () {
+  return getWindow({ index: globalThis.__args.index })
```
jwerle, line 133: this should be `  return getWindow(globalThis.__args.index)`

api/application.js
```diff
@@ -0,0 +1,303 @@
+/**
+ * @module Application
```
heapwolf, line 2: maybe we want to shorten this to `app`?

src/desktop/main.cc
```diff
@@ -652,72 +567,159 @@ MAIN {
...
+
+      JSON::Object json = JSON::Object::Entries {
+        { "data", targetWindow->json() },
+      };
+
+      currentWindow->resolvePromise(seq, OK_STATE, json.str());
+      return;
+    }
+
+    if (message.name == "window.showInspector") {
```
heapwolf, line 709: does this work on linux and windows? cc @mribbons

heapwolf (approved):
this work looks great, from what I read it looks good to merge and start trying to use it.

src/desktop/main.cc
mribbons: Inspector menu works on windows :+1: @chicoxyzzy @heapwolf

mribbons (approved):

src/core/core.cc
```diff
@@ -587,14 +587,14 @@ namespace SSC {
 
       if (!success) {
         json = JSON::Object::Entries {
-          {"source", "platform.notify"},
```
chicoxyzzy, line 590: oh, wasn't it for the OS notifications?

chicoxyzzy (commented):
👍

jwerle (approved):

src/core/core.cc
jwerle: Not there. It was a typo

PR #173, opened 2023-02-16
Windows console fix

mribbons:
windows: disable extra console

1. Added additional code to win.cc to get tests to write to console
(These are probably in the wrong place, let me know if this can be improved)
3. modified cli to use SSC::Process when launching app, so std pipes can be captured and output
4. console can be reenabled using --w32-console

### DIFF
--- a/src/app/app.cc
+++ b/src/app/app.cc
@@ -87,8 +87,10 @@ namespace SSC {
     }
 #elif defined(_WIN32)
     if (isDebugEnabled()) {
-      fclose(console);
-      FreeConsole();
+      std::cout << "PostQuitMessage(0)" << std::endl;
+      if (w32ShowConsole) {
+        HideConsole();
+      }
     }
     PostQuitMessage(0);
 #endif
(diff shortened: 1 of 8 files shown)


mribbons:
Worth mentioning:
Linux CI still failing due to double free bug, https://github.com/socketsupply/socket/pull/172 should resolve
Windows CI still hangs. I have ran this code against later code and there is no hang, let's get this and #166 merged in then see how things are.

src/process/win.cc
```diff
@@ -7,6 +7,27 @@
 
 namespace SSC {
 
+SSC::String FormatError(DWORD error, SSC::String source) {
```
heapwolf, line 10: nice, much needed

heapwolf (approved):
this seems good to go, one thing to consider is, do we want to use the standard logging function and make it output nicely to windows console? right now it spits out garbage instead of colorized ansi sequences.

src/process/win.cc
mribbons: Agree, we want to move this to a more central location though, eventually.

mribbons:
>  right now it spits out garbage instead of colorized ansi sequences.

It works now!! There was something wrong with the default terminal window, maybe it wasn't in the right mode.

src/app/app.cc
```diff
@@ -87,8 +87,10 @@ namespace SSC {
     }
 #elif defined(_WIN32)
     if (isDebugEnabled()) {
-      fclose(console);
-      FreeConsole();
+      std::cout << "PostQuitMessage(0)" << std::endl;
```
jwerle, line 90: do we need this here?

src/desktop/main.cc
```diff
@@ -142,6 +142,13 @@ MAIN {
     // launched from the `ssc` cli
     app.fromSSC = s.find("--from-ssc") == 0 ? true : false;
 
+    if (!app.w32ShowConsole) {
+      app.w32ShowConsole = s.find("--w32-console") == 0 ? true : false;
+      if (app.w32ShowConsole) {
```
jwerle, line 147: maybe this check should happen below? this `if` block?

src/window/win.cc
```diff
@@ -993,6 +993,14 @@ namespace SSC {
                           return S_OK;
                         }
 
+                        if (msg.name == "stdout") {
```
jwerle, line 996: why add this here instead of changing it in `bridge.cc` to handle windows?

src/desktop/main.cc
mribbons: Cleaned up code

src/window/win.cc
mribbons: Resolved, works via stdWrite now

jwerle (approved):

PR #180, opened 2023-02-17
w32-console: Fix output not writing to console

mribbons:


src/common.hh
```diff
@@ -547,7 +551,17 @@ namespace SSC {
   }
 
   inline void stdWrite (const String &str, bool isError) {
```
jwerle, line 553: Maybe we can drop the `#else` block and move that to the top of the function.

jwerle (approved):

PR #181, opened 2023-02-17
fix(test/src/dns.js): drop cloudflare.com

jwerle:
Random failures (`EAI_AGAIN`) on Android and Windows from cloudflare.com. I tried github.com, but I also get random failures. Maybe we should think about handling this in the test (retries?)


mribbons (approved):
Excellent, thanks 🤩

PR #184, opened 2023-02-19
windows ci/test: Resolve process status issues

mribbons:
1. Introduce SSC::Process->killed flag to prevent illegal access exception
(object was being deleted before wait thread completed)

2. Resolve exitCode -> int conversion issue:
If exitCode == 1 in previous code, (int)exitCode == 0, not good.

Note that this is basically a cross platform issue, as Win32 `GetProcessExitCode()` sets a DWORD (uint), while POSIX like `wait_pid` sets an int_32_t (currently on macos).

### DIFF
--- a/src/desktop/main.cc
+++ b/src/desktop/main.cc
@@ -213,6 +213,7 @@ MAIN {
     if (processToKill != nullptr) {
       auto pid = processToKill->getPID();
       processToKill->kill(pid);
+      processToKill->wait();
 
       if (processToKill == process) {
         process = nullptr;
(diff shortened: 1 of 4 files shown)


heapwolf (approved):
We could definitely use a test around this in the javascript apis /cc @chicoxyzzy

chicoxyzzy (commented):
can we reuse the status unstead of introducing a new atomic? cc @jwerle

mribbons:
> can we reuse the status unstead of introducing a new atomic? cc @jwerle

Great idea @chicoxyzzy 
Seems to work on windows, have force pushed to remove killed status.
Hopefully CI passes on linux/mac.

mribbons:
@chicoxyzzy Success!

trevnorris (approved):

PR #189, opened 2023-02-21
ios logs

chicoxyzzy:


src/cli/cli.cc
```diff
@@ -383,6 +383,18 @@ void runIOSSimulator (const fs::path& path, Map& settings) {
     exit(rInstallApp.exitCode);
   }
 
+  log("launching the child logger process");
```
jwerle, line 386: maybe we output this for debug only?

chicoxyzzy: Yes, I don't think we need it. The purpose of this PR is to save a draft of how to enable logging. Currently, streaming doesn't work for me, I only see a first message from `log stream` about starting to listen to the log

chicoxyzzy:
there's also a `xcrun simctl spawn booted log stream` command which I believe will only watch for the booted device

src/cli/cli.cc
jwerle: yep log stream only works for local stores

PR #190, opened 2023-02-21
Diagnostics API

jwerle:
This pull request introduces an initial _Diagnostics API_ for usable in user space and also suitable for retrieving real time information about the usage of various API modules.

**listening for UDP sockets created and counting bytes of sent/received network traffic:**

```js
import diagnostics from 'socket:diagnostics'

const group = diagnostics.channels.group('udp')

group.subscribe('socket', (message) => {
  const { socket } = message // `socket` is an instance of `dgram.Socket`
})

let bytesReceived = 0
group.subscribe('message', (message) => {
  const { buffer } = message
  bytesReceived += buffer.length
})

let bytesWritten = 0
group.subscribe('send', (message) => {
  const { buffer } = message
  bytesWritten += buffer.length
})
```

**tracking requestAnimationFrame rate:**

```js
import diagnostics from 'socket:diagnostics'

diagnostics.window.metrics.start(['requestAnimationFrame'])
diagnostics.window.metrics.subscribe('requestAnimationFrame', (message) => {
  console.log(message.rate) // current rate of usage for `requestAnimationFrame()`
})

// start loop and sample 64 frames
let iterations = 0
requestAnimationFrame(function loop () {
  if (iterations++ < 64) {
    requestAnimationFrame(loop)
  }
})

diagnostics.window.metrics.stop(['requestAnimationFrame'])
```

**tracking `fetch()` calls and `XMLHttpRequest` usage:**

```js
import diagnostics from 'socket:diagnostics'

diagnostics.window.metrics.start(['fetch', 'XMLHttpRequest'])

diagnostics.window.metrics.subscribe('fetch', (message) => {
  console.log(message.resource) // the first argument to `fetch()` 
})

diagnostics.window.metrics.subscribe('XMLHttpRequest.*', (message, metric) => {
  if (metric.name === 'window.XMLHttpRequest.send') {
    const { body } = message
  }

  if (metric.name === 'window.XMLHttpRequest.open') {
    const { method, url } = message
  }
})

diagnostics.window.metrics.stop(['fetch', 'XMLHttpRequest'])
```

### DIFF
--- a/api/bluetooth.js
+++ b/api/bluetooth.js
@@ -4,10 +4,22 @@
  * A high-level, cross-platform API for Bluetooth Pub-Sub
  */
 
-import * as ipc from './ipc.js'
 import { EventEmitter } from './events.js'
+import diagnostics from './diagnostics.js'
+import ipc from './ipc.js'
 
 import * as exports from './bluetooth.js'
+
+export default exports
+
+const dc = diagnostics.channels.group('bluetooth', [
+  'data',
+  'event',
+  'start',
+  'handle',
+  'publish',
+  'subscribe'
+])
 
 /**
  * Create an instance of a Bluetooth service.
@@ -37,6 +49,12 @@ export class Bluetooth extends EventEmitter {
 
       if (data?.serviceId === this.serviceId) {
         this.emit(data.characteristicId, data, e.detail.data)
+        dc.channel('data').publish({
+          bluetooth: this,
+          serviceId,
+          characteristicId: data.characteristicId,
+          data: e.detail.data
+        })
       }
     })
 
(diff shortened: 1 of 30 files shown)


mribbons (approved):
My main thought here is that the Diagnostics code is tightly coupled with each of the areas that require metrics.

Could we use an IOC or event based approach instead?
Obviously there's going to be domain specific metrics that need to be measured, but in essence these would all just be counters or moving averages (eventually).

I know it's late in the dev cycle for this kind of change :)

heapwolf (approved):

jwerle:
> My main thought here is that the Diagnostics code is tightly coupled with each of the areas that require metrics.
> 
> Could we use an IOC or event based approach instead? Obviously there's going to be domain specific metrics that need to be measured, but in essence these would all just be counters or moving averages (eventually).
> 
> I know it's late in the dev cycle for this kind of change :)

In this case, we'll avoid [IOC](https://en.wikipedia.org/wiki/Inversion_of_control) as we want as thing of a layer between the built in diagnostic channels and the subscribers. These paths could be [_hot_](https://en.wikipedia.org/wiki/Hot_spot_(computer_programming)). 

From the outside, the diagnostics API is the inversion of control, with an "event" driven model built in to the public API

chicoxyzzy (approved):

PR #200, opened 2023-02-23
Launch macOS application with NSWorkspace and use OSLogStore for log retrieval

jwerle:
This was a wild one, but I tried to leave as many comments as possible!

### DIFF
--- a/test/src/process.js
+++ b/test/src/process.js
@@ -3,9 +3,6 @@ import process from 'socket:process'
 import { primordials } from 'socket:ipc'
 import path from 'path-browserify'
 
-// make `path-browserify` happy
-globalThis.process = process
-
 test('process', (t) => {
   t.ok(typeof process.addListener === 'function', 'process is an EventEmitter')
 })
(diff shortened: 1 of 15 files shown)


chicoxyzzy:
I see this in my terminal
```
# fs.watch
# fs.write
# fs.writeFile
2023-02-24 05:16:37.711 ssc[66872:8775122] error: OSLogStore: The log archive is corrupt or incomplete and cannot be read
```
and then I don't see any log output

jwerle:
> I see this in my terminal
> 
> ```
> # fs.watch
> # fs.write
> # fs.writeFile
> 2023-02-24 05:16:37.711 ssc[66872:8775122] error: OSLogStore: The log archive is corrupt or incomplete and cannot be read
> ```
> 
> and then I don't see any log output

It should be fixed now!

mribbons:
I'm just going to say: It's a lot of platform specific code 😄

chicoxyzzy (approved):
it works for me now

chicoxyzzy:
BTW @jwerle maybe you can take a look at https://github.com/socketsupply/socket/pull/189 as well? I'm stuck there

test/src/process.js
```diff
@@ -3,6 +3,9 @@ import process from 'socket:process'
 import { primordials } from 'socket:ipc'
 import path from 'path-browserify'
 
+// make `path-browserify` happy
+globalThis.process = process
```
chicoxyzzy, line 7: We can also delete it from `globalThis` at the and of the module

chicoxyzzy: actually, we need `path.resolve` in the `socket:path` :)

jwerle: haha yes TODO!

jwerle:
> BTW @jwerle maybe you can take a look at #189 as well? I'm stuck there

we'll have to think about our approach here as using `OSLogStore` won't work

PR #201, opened 2023-02-23
Window configuration

chicoxyzzy:


chicoxyzzy:
I only found how to set a the minimum width/height on macOS, no max width/height on macOS and no min/max width/height on other OSes

jwerle (approved):

PR #202, opened 2023-02-24
Socket win32 x64 0.0.4

mribbons:


mribbons:
This PR also covers 0.0.2 and 0.0.4...

npm/packages/@socketsupply/socket-win32-x64/bin/install-pre-reqs.js
```diff
@@ -1,9 +1,59 @@
...
+
+  console.log(envs);
+
+  fs.writeFileSync("env.bat", envs);
+
+  var ssc_cmd_path = path.join(process.cwd(), '../../.bin/ssc.cmd');
+
+
+  if (!fs.existsSync(ssc_cmd_path)) {
+    console.log(`WARNING: No ${ssc_cmd_path} builds will fail - please report (https://discord.gg/YPV32gKCsH)`);
```
jwerle, line 33:
```js
WARNING: Unable to build - ${ssc_cmd_path} is missing - please report (https://discord.gg/YPV32gKCsH)
```

cc @heapwolf what messaging do you think should be here?

bin/install.ps1
```diff
@@ -398,4 +409,15 @@ if ($global:path_advice.Count -gt 0) {
...
 
+if (-not $shbuild) {
+  Write-Output "Please close this terminal to continue setting up your socket app."
+}
+
+if ($package_setup -eq $true) {
+  $paths = @{}
+  $fso = New-Object -ComObject Scripting.FileSystemObject 
+  $paths["CXX"] = $fso.GetFile($(Get-CommandPath "clang++.exe")).ShortPath
+  ConvertTo-Json $paths > env.json
```
jwerle, line 420: maybe we can move this whole block here to a separate file that is used to for this setup in npm

mribbons: @jwerle It needs to be in install.ps1 because `vcvars64.bat` has been executed so `clang++` is in the path. After npm setup completes, the environment is binned and `clang++` is lost until `vcvars64.bat` runs again.

PR #207, opened 2023-02-26
Handle signals in 'ssc(1)' to shutdown child app process

jwerle:
We were no longer handling signals with the move to `SSC::Process` from `SSC::exec`. This pull request introduces a signal handler.

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -148,11 +148,11 @@ static std::atomic<int> appStatus = 0;
 static std::mutex appMutex;
 
 void signalHandler (int signal) {
+  appStatus = signal;
+
   if (appProcess != nullptr) {
     auto pid = appProcess->getPID();
     appProcess->kill(pid);
-    delete appProcess;
-    appProcess = nullptr;
   } else if (appPid > 0) {
     kill(appPid, signal);
     appPid = 0;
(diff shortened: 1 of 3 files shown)


src/cli/cli.cc
```diff
@@ -142,6 +142,23 @@ inline String prefixFile () {
...
+static Process* appProcess = nullptr;
+
+void signalHandler (int signal) {
+  if (appProcess != nullptr) {
+    auto pid = appProcess->getPID();
+    appProcess->kill(pid);
+    delete appProcess;
+    appProcess = nullptr;
+  } else if (appPid > 0) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(32));
```
chicoxyzzy, line 155: nit: indentation

chicoxyzzy (approved):

src/cli/cli.cc
chicoxyzzy: nvm it was a GitHub glitch

src/cli/cli.cc
```diff
@@ -142,9 +142,27 @@ inline String prefixFile () {
...
+static std::atomic<int> appStatus = 0;
+static std::mutex appMutex;
+
+void signalHandler (int signal) {
+  appStatus = signal;
+
+  if (appProcess != nullptr) {
+    auto pid = appProcess->getPID();
+    appProcess->kill(pid);
+  } else if (appPid > 0) {
```
mribbons, line 156: WIN32 doesn't have `kill` in `signal.h`, we need the following:
```cpp
socket/src/cli/cli.cc:157:5: error: use of undeclared identifier 'kill'
    kill(appPid, signal);
    ^
```

```cpp
#ifndef _WIN32
  else if (appPid > 0) {
    kill(appPid, signal);
    appPid = 0;
  } 
#endif
```

mribbons (changes requested):

mribbons:
Aside from the change I requested, this PR works great @jwerle!

mribbons (approved):
Excellent :tada:

PR #208, opened 2023-02-27
Various fixes

jwerle:
- Assigns `libsodium` export to `crypto.sodium` export 
- Falls back to `sodium.randombytes_buf` in `crypto.getRandomValues`
- Exports `CustomEvent` in `socket:events` for usage in platforms that do not support it globally (node)
- Allow `encoding` to be given in `util.toBuffer()` where `Buffer.from()` is used
- Use `value` as `this` value when calling custom inspect (`Symbol.for('socket.util.inspect.custom')`)
- Introduce `compareBuffers()` for comparing `all `TypedArray` types, including `Buffer`
- Support `--link` for `publish-npm-modules.sh` which globally links the built node modules, instead of publishing  them (this makes it easier to develop against: `npm link @socketsupply/socket` just works)

### DIFF
--- a/api/events.js
+++ b/api/events.js
@@ -469,6 +469,21 @@ function eventTargetAgnosticAddListener (emitter, name, listener, flags) {
 EventEmitter.EventEmitter = EventEmitter
 EventEmitter.once = once
 
+export const CustomEvent = globalThis.CustomEvent || class CustomEvent extends globalThis.Event {
+  #detail = null
+
+  constructor (type, options) {
+    super(type, options)
+    if (options?.detail) {
+      this.#detail = options.detail
+    }
+  }
+
+  get detail () {
+    return this.#detail
+  }
+}
+
 export {
   EventEmitter,
   once
--- a/api/crypto.js
+++ b/api/crypto.js
@@ -5,6 +5,7 @@
  * random bytes and hashing.
  */
 
+import { toBuffer } from './util.js'
 import { Buffer } from './buffer.js'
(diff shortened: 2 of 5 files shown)


mribbons (approved):

api/events.js
```diff
@@ -469,6 +469,21 @@ function eventTargetAgnosticAddListener (emitter, name, listener, flags) {
 EventEmitter.EventEmitter = EventEmitter
 EventEmitter.once = once
 
+export const CustomEvent = globalThis.CustomEvent || class CustomEvent extends globalThis.Event {
```
chicoxyzzy, line 472: is this for Node.js?

chicoxyzzy (approved):
it nice to have documentation for npm publishing somewhere (in wiki?)

api/events.js
jwerle: yep!

jwerle:
> it nice to have documentation for npm publishing somewhere (in wiki?)

yeah I agree 🙏🏼  here is some inline until we get a wiki setup on the repo:

```sh
# '--link' will perform a `npm link` and also link the platforms in the top level (it implies '--dry-run' so `npm publish` is not executed
# '--dry-run' does everything except `npm publish`
# '--only-platforms' only publishes (or links) platform modules
# '--only-top-level' only publishes (or links) top level module
publish-npm-modules.sh [--link] [--dry-run] [--only-platforms] [--only-top-level]
```

PR #211, opened 2023-02-28
Fix debug logs on macOS

jwerle:
We were not handling debug output to the `OSLogStore`. This properly handles debug logs for a macos app and queries them in the `ssc` cli


chicoxyzzy (approved):
works fine on my machine!

PR #214, opened 2023-03-01
Windows socket scheme + Win32 debug build fix

mribbons:
Win32 socket: scheme is partially implemented, I had to resolve the below issue first.

install.sh: fix Win32 debug

Problem:
When building libuv and socket-runtime with debug enabled, our apps crash when 
using ifstream (readFile):
Debug Assertion Failed. Expression: (_osfile(fh) & fopen)
This build issue also interferred with debugging in Visual Studio.

This occured because by default the clang incorrectly links to the non 
threaded, production version of C runtime .lib (libcrt)
After taking the nessary steps to manually link to the correct lib 
(Including adding preprocessor definitions for /MT[d]),
ssc and apps would not link, therefore:

Solution:
This commit splits debug and release build artifacts:
d is set if $DEBUG and $host == Win32
*.o files are now named *$d.o
Libs are copied to build/platform/lib$d
libsocket-runtime.a is now named libsocket-runtime$d.a
ssc build --prod defines whether or not the app is being built for debug 
and therefore links to the correct libsocket-runtime$d.a

There should be no changes on other OS's.

More changes are required:
install.ps1 will need to build a WIN_DEBUG_LIBS and inject into ssc.cmd
publish-npm-modules.sh should ship release and debug libs on windows
(Call install.sh with DEBUG enabled and disabled)

### DIFF
--- a/test/build.js
+++ b/test/build.js
@@ -1,11 +1,8 @@
 #!/usr/bin/env node
 import path from 'node:path'
 import fs from 'node:fs/promises'
-import os from 'node:os'
 
 import esbuild from 'esbuild'
-
-const dirname = path.resolve(path.dirname(import.meta.url.replace('file://', '').replace(/^\/[A-Za-z]:\//, '/')))
 
 const cp = async (a, b) => fs.cp(
   path.resolve(a),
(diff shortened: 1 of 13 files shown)


test/build.js
```diff
@@ -35,25 +32,25 @@ async function main () {
...
-      setup (build) {
-        build.onResolve({ filter: /^socket:.*$/ }, (args) => {
-          const basename = args.path.replace('socket:', '').replace(/.js$/, '') + '.js'
-          const filename = path.resolve(path.dirname(dirname), 'api', basename)
-          return { path: filename, external: true }
-        })
-      }
-    })
-  }
+  // if (os.platform() !== 'win32') {
```
jwerle, line 40: maybe a comment above explaining why this is commented out if we're going to leave it 🙏🏼

jwerle (commented):
LGTM!

jwerle (approved):

PR #216, opened 2023-03-02
Modules clean up

jwerle:
- Drops import maps from Android as we handle the `socket:` protocol in request interception.
- Cleans up some of `api/`
- Ensure `socket/` directory copied after user build command is invoked


chicoxyzzy (approved):
lgtm!

PR #218, opened 2023-03-03
Introduce 'os.hrtime()' and 'os.availableMemory()'

jwerle:
This PR also adds a few more missing porcelain functions for `os` and `process`

### DIFF
--- a/src/desktop/main.cc
+++ b/src/desktop/main.cc
@@ -499,6 +499,11 @@ MAIN {
       } catch (...) {
       }
 
+    #if defined(__APPLE__)
+      if (app.fromSSC) {
+        debug("__EXIT_SIGNAL__=%d", exitCode);
+      }
+    #endif
       window->exit(exitCode);
       return;
     }
--- a/api/buffer.js
+++ b/api/buffer.js
@@ -352,6 +352,10 @@ function createBuffer (length) {
  * The `Uint8Array` prototype remains unmodified.
  */
 
+/**
+ * @name Buffer
+ * @extends {Uint8Array}
+ */
 function Buffer (arg, encodingOrOffset, length) {
   // Common case.
   if (typeof arg === 'number') {
--- a/api/crypto.js
+++ b/api/crypto.js
@@ -53,8 +53,8 @@ export function getRandomValues (buffer, ...args) {
     return webcrypto?.getRandomValues(buffer, ...args)
   }
 
-  if (sodium.libsodium.randombytes_buf) {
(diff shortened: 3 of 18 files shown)


mribbons:
@jwerle looks good. Is the windows CI issue caused by log poll changes, or is it due to new code?

mribbons:
@jwerle just tried this on my Win32, I get the same result as CI:
```
# os.hrtime()
Uncaught InvalidAccessError: Failed to execute 'open' on 'XMLHttpRequest': Synchronous requests from a document must not set a response type.

caught InvalidAccessError: Failed to execute 'open' on 'XMLHttpRequest': Synchronous requests from a document must not set a response type.
WARNING: Window#0 exiting with code 1
execute 'open' on 'XMLHttpRequest': Synchronous requests from a document must not set a response type.
```

jwerle:
> @jwerle just tried this on my Win32, I get the same result as CI:
> 
> ```
> # os.hrtime()
> Uncaught InvalidAccessError: Failed to execute 'open' on 'XMLHttpRequest': Synchronous requests from a document must not set a response type.
> 
> caught InvalidAccessError: Failed to execute 'open' on 'XMLHttpRequest': Synchronous requests from a document must not set a response type.
> WARNING: Window#0 exiting with code 1
> execute 'open' on 'XMLHttpRequest': Synchronous requests from a document must not set a response type.
> ```

wow! TIL. Okay I can work around this!

chicoxyzzy (approved):

jwerle:
> @jwerle just tried this on my Win32, I get the same result as CI:
> 
> ```
> # os.hrtime()
> Uncaught InvalidAccessError: Failed to execute 'open' on 'XMLHttpRequest': Synchronous requests from a document must not set a response type.
> 
> caught InvalidAccessError: Failed to execute 'open' on 'XMLHttpRequest': Synchronous requests from a document must not set a response type.
> WARNING: Window#0 exiting with code 1
> execute 'open' on 'XMLHttpRequest': Synchronous requests from a document must not set a response type.
> ```

okay this _should_ be fixed!

src/desktop/main.cc
```diff
@@ -499,6 +499,11 @@ MAIN {
       } catch (...) {
       }
 
+    #if defined(__APPLE__)
+      if (app.fromSSC) {
```
jwerle, line 503: this is only on macOS desktop and when the application is launched from the `ssc` CLI. This allows us to get the exit code of the process, which is impossible from `ssc` as we launch the application with `NSWorkspace` and the actual process that gets launched is **NOT** a child process. So `wait`, `waitpid` and `waitid` will _NOT_ work here :(

PR #219, opened 2023-03-03
socket-win32-x64: better install process

mribbons:
%comspec% call is no long required, sub process can run in the same session
(Due to the fact that install-pre-reqs will
now have access to stdin)

Requires create-socket-app@1.0.13

### DIFF
--- a/bin/install.ps1
+++ b/bin/install.ps1
@@ -80,6 +80,7 @@ Function Test-CommandVersion {
   $r = $($null -eq $F.length)
   if ($r -eq $false) {
     Write-Output $r
+    return
   }
 
   $output = iex "& $command_string --version" | Out-String
--- a/bin/install.ps1
+++ b/bin/install.ps1
@@ -25,6 +25,7 @@ $global:useCurl = $true
 $global:WIN_DEBUG_LIBS = ""
 $global:path_advice = @()
 $global:install_errors = @()
+$targetClangVersion = "15.0.0"
 
 Function Exit-IfErrors {
   if ($global:install_errors.Count -gt 0) {
@@ -65,7 +66,7 @@ $vsconfig = "nmake.vsconfig"
 
 if ( -not (("llvm+vsbuild" -eq $toolchain) -or ("vsbuild" -eq $toolchain) -or ("llvm" -eq $toolchain)) ) {
   Write-Output "Unsupported -toolchain $toolchain. Supported options are vsbuild, llvm+vsbuild or llvm (external nmake required)"
-  Write-Output "-toolchain llvm+vsbuild will check for and install llvm clang and vsbuild nmake."
+  Write-Output "-toolchain llvm+vsbuild will check for and install llvm clang $targetClangVersion and vsbuild nmake."
   
   Exit 1
 }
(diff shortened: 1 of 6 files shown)


bin/install.ps1
```diff
@@ -330,40 +430,56 @@ Function Install-Requirements {
     if ($vc_exists) {
       $(Get-ProcEnvs($vc_vars))
     } else {
-      Write-Output "vcvars64.bat still not present, something went wrong."
-      Exit 1
+      if ($env:CI -eq $null) {
+        $global:install_errors += "vcvars64.bat still not present, something went wrong."
```
heapwolf, line 434: should this bail here?

npm/packages/@socketsupply/socket-win32-x64/bin/install-pre-reqs.js
```diff
@@ -16,40 +16,50 @@ async function insertEnvVars () {
...
+        sscCmdData = sscCmdData.replace('@ECHO off\r\n', `@ECHO off\r\n${setLine}`)
+        updated = true
+      }
+    }
   }
 
-  console.log(envs)
-
   fs.writeFileSync('env.bat', envs)
+  if (updated === true) { fs.writeFileSync(sscCmdPath, sscCmdData) }
```
heapwolf, line 47: does `fs.writeFileSync` this need a try catch?

npm/packages/@socketsupply/socket-win32-x64/bin/install-pre-reqs.js
```diff
@@ -16,40 +16,50 @@ async function insertEnvVars () {
...
 
-  const sscCmdPath = path.join(process.cwd(), '../../.bin/ssc.cmd')
-
-  if (!fs.existsSync(sscCmdPath)) {
-    console.log(`WARNING: No ${sscCmdPath} builds will fail - please report (https://discord.gg/YPV32gKCsH)`)
-    return
+  if (fs.existsSync('ps1/ps1_errors.txt')) {
+    throw fs.readFileSync('ps1/ps1_errors.txt', 'utf-8')
```
heapwolf, line 50: should this `throw new Error('description')`?

heapwolf (approved):

bin/install.ps1
mribbons: This function gets called at the end of the script. It allows us to report multiple issues instead of failing on a single issue, as well as only return a bad status code depending on the context of how the script was called.

https://github.com/socketsupply/socket/pull/219/files#diff-e9a96fb537f64b60e955f6d260d0d3c697cb45050752cf4d834833486b4d8530R46-R53

npm/packages/@socketsupply/socket-win32-x64/bin/install-pre-reqs.js
mribbons: I think it's fine as is - It's running in the context of the npm install process, the location has already been confirmed by this point.

npm/packages/@socketsupply/socket-win32-x64/bin/install-pre-reqs.js
mribbons: This one isn't required since we run install-pre-reqs directly from `create-socket-app`'s `index.js` (rather than in `preinstall`, `postinstall`)

jwerle (approved):

PR #222, opened 2023-03-07
Preload Refactor

jwerle:
Imports `socket:init` in preload script for global setup, like managing the concurrency of `ipc://post` XHR requests. Heavy usage makes the webview freeze and these are uninitiated requests, so we should throttle them. This PR also makes the runtime work in web workers


chicoxyzzy:
hmm, maybe we can initialize primordials in the same phase?

chicoxyzzy:
and FS constants

jwerle:
> hmm, maybe we can initialize primordials in the same phase?

they are with `import ipc from 'socket:ipc'`

jwerle:
@chicoxyzzy this one is ready to go!

src/core/core.cc
```diff
@@ -109,41 +109,28 @@ namespace SSC {
...
-      "xhr.send();                                                   \n"
+      "const globals = await import('socket:internal/globals');              \n"
+      "const id = `" + sid + "`;                                             \n"
+      "const seq = `" + seq + "`;                                            \n"
+      "const headers = `" + trim(post.headers) + "`                          \n"
+      "  .trim()                                                             \n"
+      "  .split(/[\\r\\n]+/)                                                 \n"
+      "  .filter(Boolean);                                                   \n"
+      "                                                                      \n"
+      "let params = `" + params + "`;                                        \n"
```
heapwolf, line 120: does this need to escape backticks?

heapwolf (approved):

src/core/core.cc
jwerle: it shouldn't! I just copied this from what was there before

mribbons (changes requested):
I know it's not related to this branch specifically, but:
```cli.cc:768:10: error: use of undeclared identifier 'SIGUSR1'```

Also, fix linting :)

chicoxyzzy (approved):
LGTM

PR #223, opened 2023-03-07
Android prebuild

mribbons:
Android prebuilds are built by default, as long as `ANDROID_HOME` is set.

See notes in https://github.com/socketsupply/socket/blob/android-prebuild/bin/build-android-runtime-libraries.sh

CI:
`ANDROID_HOME` seems to be set in all desktop variants.
Building Android for these slows things down too much, therefore we only build android if
`SSC_ANDROID_CI` is set.

### DIFF
--- a/src/cli/templates.hh
+++ b/src/cli/templates.hh
@@ -1008,7 +1008,7 @@ android {
       abiFilters {{android_ndk_abi_filters}}
     }
 
-{{default_config_external_native_build}}
+{{android_default_config_external_native_build}}
   }
 
   aaptOptions {
--- a/src/cli/templates.hh
+++ b/src/cli/templates.hh
@@ -1008,7 +1008,7 @@ android {
       abiFilters {{android_ndk_abi_filters}}
     }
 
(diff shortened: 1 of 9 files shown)


src/cli/templates.hh
```diff
@@ -1422,6 +1415,7 @@ version = 1.0.0
 
 
 [android]
+build_remove_path = dist/android/app/src/
```
chicoxyzzy, line 1418: can we use the `settings["output"] + /android/app/src/` in cli.cc to not introduce extra config parameters?

heapwolf: +1 if this is possible, I do this in some other places

mribbons: Seems to work fine, all done.

chicoxyzzy (approved):
I don't have an Android device to test on, but the code looks good to me

chicoxyzzy:
hmm macOS tests are failing

jwerle:
@mribbons can we close this one in favor of #234 ?

PR #226, opened 2023-03-07
provide 'fetch()' response clone in metrics, and introduce 'WorkerMetric'

jwerle:


api/diagnostics/window.js
```diff
@@ -175,10 +176,37 @@ export class XMLHttpRequestMetric extends Metric {
...
 }
 
+export class WorkerMetric extends Metric {
+  static GlobalWorker = globalThis.Worker
+
+  constructor (options) {
+    super()
+    const self = this
+    this.channel = dc.channel('Worker')
+    this.Worker = class Worker extends globalThis.Worker {
```
heapwolf, line 186: 😎

heapwolf (approved):

mribbons (changes requested):
I know it's not related to this branch specifically, but:
cli.cc:768:10: error: use of undeclared identifier 'SIGUSR1'

Also, fix linting :)

chicoxyzzy (approved):
👍

api/diagnostics/window.js
```diff
@@ -175,10 +176,37 @@ export class XMLHttpRequestMetric extends Metric {
   }
 }
 
+export class WorkerMetric extends Metric {
+  static GlobalWorker = globalThis.Worker
+
+  constructor (options) {
+    super()
+    const self = this
```
chicoxyzzy, line 184: linter is not happy that `self` is not used after the variable initialization

PR #229, opened 2023-03-09
Fix Android threading

jwerle:
Given enough concurrent load, the UI thread locks up. This PR introduces concurrency control and moves route processing to a thread to prevent request interception from deadlocking or the UI thread from blocking

### DIFF
--- a/api/internal/init.js
+++ b/api/internal/init.js
@@ -1,5 +1,6 @@
 /* global CustomEvent, Blob */
 /* eslint-disable import/first */
+// mark when runtime did init
 globalThis.__RUNTIME_INIT_NOW__ = performance.now()
 
 import { IllegalConstructor, InvertedPromise } from '../util.js'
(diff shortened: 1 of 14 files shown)


api/internal/init.js
```diff
@@ -21,105 +32,165 @@ class RuntimeXHRPostQueue extends EventTarget {
...
-          return this.onmessage(event)
+class RuntimeWorker extends GlobalWorker {
+  #onmessage = null
+
+  constructor (url, options, ...args) {
+    const preload = `
+    globalThis.__args = ${JSON.stringify(globalThis.__args)}
+    await import('socket:internal/init')
+    await import('socket:internal/worker')
+    await import('${new URL(url, globalThis.location?.href || '/')}')
```
chicoxyzzy, line 93: is the order necessary here? may some of these imports work in parallel?

heapwolf (approved):

mribbons:
Ok, tested with @heapwolf's relay app, gets around the startup freeze.

mribbons (approved):

PR #234, opened 2023-03-11
Android static runtime libs

mribbons:
`build-runtime-library.sh` builds `libsocket-runtime.a`
`install.sh` builds `uv.a`

Unlike https://github.com/socketsupply/socket/pull/223, we don't require ssc to be built in order to build the static libs.

No Android makefiles are required, except for building the final app.

`__BUNDLE_IDENTIFIER__` wasn't an issue because all `android/*.cc` files are built during app build phase.


src/cli/templates.hh
```diff
@@ -1092,58 +1086,19 @@ kotlin.code.style=official
 constexpr auto gAndroidMakefile = R"MAKE(
 LOCAL_PATH := $(call my-dir)
 
-## libuv.a
+## uv.a
```
jwerle, line 1089: can we maintain `libuv.a` here as it is the correct library name 🙏🏼 
we often `git grep libuv` or `ag libuv` and this would be missed in that

src/cli/cli.cc
```diff
@@ -2271,139 +2392,166 @@ int main (const int argc, const char* argv[]) {
...
+          {        
+            log(ndkBuildArgs.str());
+            log("ndk build failed.");
+            exit(1);
+          }
+        }
+      }
+
+      // just build for CI
+      if (getEnv("SSC_CI").size() > 0) {
```
jwerle, line 2516: 👍🏼

src/cli/cli.cc
```diff
@@ -2429,28 +2578,45 @@ int main (const int argc, const char* argv[]) {
...
+          adb << androidHome << "/platform-tools/";
+        } else {
           adb << androidHome << "\\platform-tools\\";
+        }
+        
+        if (debugEnv || verboseEnv) log((adb.str() + (" --version > ") + SSC::String((!platform.win) ? "/dev/null" : "NUL") + (" 2>&1")));
+        if (!std::system((adb.str() + (" --version > ") + SSC::String((!platform.win) ? "/dev/null" : "NUL") + (" 2>&1")).c_str())) {
+          log("Warn: Failed to locate adb at " + adb.str());
+        } else {
+          log("got adb at " + adb.str());
```
jwerle, line 2593: seems like debug output to me

src/cli/templates.hh
mribbons: Good call.

bin/build-runtime-library.sh
```diff
@@ -225,14 +215,24 @@ function main () {
...
       echo "ok - using cached static library ($arch-$platform): $(basename "$static_library")"
     else
       echo "static library doesn't exist after cache check passed: ($arch-$platform): $(basename "$static_library")"
+      exit 1
     fi
   fi
+
+  lib_size=$(stat_size $static_library)
+
+  if (( $lib_size < 1048576 )); then
```
jwerle, line 231: Why are we checking for the size? Why not just bail when the library fails to build or doesn't exist

bin/install.sh
```diff
@@ -565,6 +571,90 @@ function _setSDKVersion {
...
+      echo "ok - using cached static library ($arch-$platform): $(basename "$static_library")"
+    else
+      echo "static library doesn't exist after cache check passed: ($arch-$platform): $(basename "$static_library")"
+      exit 1
+    fi
+  fi
+
+  lib_size=$(stat_size $static_library)
+
+  if (( $lib_size < 800000 )); then
```
jwerle, line 651: same here, this check seems like it may be too pedantic

bin/publish-npm-modules.sh
```diff
@@ -63,15 +78,27 @@ mkdir -p "$SOCKET_HOME"
...
+    exit 1
+  fi
 fi
 
-if (( !only_top_level )); then
+if (( !only_top_level ))  && (( !no_rebuild )) ; then
   "$root/bin/install.sh" || exit $?
 fi
 
+if [[ "$(find $SOCKET_HOME -name lib*.a | grep android | wc -l)" -lt "8" ]]; then
```
jwerle, line 96: This seems to brittle - why are we checking lib count? This is also special treatment for android

bin/publish-npm-modules.sh
```diff
@@ -106,6 +133,11 @@ if (( !only_top_level )); then
     cp -rf "$SOCKET_HOME/src"/* "$SOCKET_HOME/packages/$package/src"
     cp -rf "$SOCKET_HOME/include"/* "$SOCKET_HOME/packages/$package/include"
 
+    # don't copy debug files, too large
+    rm -rf $SOCKET_HOME/lib/*-android/objs-debug 
+    cp -rf $SOCKET_HOME/lib/*-android "$SOCKET_HOME/packages/$package/lib"
+    # cp -f $SOCKET_HOME/lib/*-android/*.a "$SOCKET_HOME/packages/$package/lib"
```
jwerle, line 139: probably should just remove this line if it is not used

src/cli/cli.cc
```diff
@@ -2271,139 +2392,166 @@ int main (const int argc, const char* argv[]) {
...
-        log("You may need to add accept_sdk_licenses = \"y\" to the [android] section of socket.ini.");
         exit(1);
       }
 
-      if (flagDebugMode) {
-        gradlew
-          << localDirPrefix
-          << "gradlew :app:bundleDebug "
-          << "--warning-mode all ";
+      if (!android_enable_standard_ndk_build)
```
jwerle, line 2430: curly

src/cli/cli.cc
```diff
@@ -1409,6 +1473,12 @@ int main (const int argc, const char* argv[]) {
       writeFile(paths.pathResourcesRelativeToUserBuild / "Credits.html", credits);
     }
 
+
+    // used in multiple if blocks, need to declare here
+    auto android_enable_standard_ndk_build = settings["android_enable_standard_ndk_build"] == "true";
```
jwerle, line 1478: ideally, camel casing is preserved throughout the file for all identifiers and functions

src/cli/cli.cc
```diff
@@ -144,6 +144,57 @@ String getSocketHome () {
...
+      }
+    }
+  }
+
+  if (androidHome.size() == 0) {
+    if (platform.mac) {
+      androidHome = getEnv("HOME") + "/Library/Android/sdk";
+    } else if (platform.unix) {
+      androidHome = getEnv("HOME") + "/android";
+    } else if (platform.win) {
```
jwerle, line 180: I know I left this as a `TODO`, but perhaps there is a standard location where the Android SDK is on Windows like `%localappdata%\Android\Sdk` (aka `C:\Users\user\AppData\Local\Android\Sdk`)

bin/publish-npm-modules.sh
```diff
@@ -63,15 +78,27 @@ mkdir -p "$SOCKET_HOME"
 export SOCKET_HOME
 export PREFIX
 
-if (( do_global_link && !dry_run )); then
-  echo >&2 "warning: '--link' implies '--dry-run' too"
-  dry_run=1
+# Confirm that ssc in PATH matches current commit
```
jwerle, line 81: While I like the intent here, I am not a fan of this type of check existing. There are edge cases to consider, like an already installed version of `ssc` that can be resolved within your `PATH` that takes precedence over the one installed from this script.

chicoxyzzy (approved):
Code lgtm, but I don't really know what happens in some parts of this patch and don't have an Android or Android emulator set up. So I think it's better to wait for someone to test this

jwerle (changes requested):

bin/build-runtime-library.sh
mribbons: It's just a fail safe.

I  have seen `llvm-ar` create 8 byte files when `${objects[@]}` was empty.

Sure we could check the array, but a better sanity check is having a decent number of bytes.
We could go further, like dumping the objects from the .a and checking vs the c source, but we're not writing a compiler.
If 1024^2 ends up being too small, the responsible dev can adjust at the time.

I have implemented an exists check in `publish-npm-modules.sh`.

bin/publish-npm-modules.sh
mribbons: I implemented the check for android because:
1. There's more effort involved in setting it up
2. ios libs are pretty much out of the box, so there's a low probability they won't build
3. You can't do much without `ssc` building, so unlikely that someone would try to ship a package without desktop libs and `ssc` bin.
4. Android adds considerably more workload to the build. I envisage switching off android builds (and IOS eventually) for optimised build runs catered to the current build target.

If I had implemented an option for IOS libs being excluded from build, I would have also added this check.

You raise a good point, I'm all for implementing these checks across the board.

src/cli/cli.cc
mribbons: Next PR.

bin/publish-npm-modules.sh
mribbons: As above, my intention is to ensure a sane package before publishing.

It's a bit of extra work to go through but I think it's well worth it.

What if instead of checking the `PATH` version, we check the instance that is copied to the npm staging folder?

bin/install.sh
mribbons: See above. libuv.a is just over 800000 bytes for 2 of the android abis, the other 2 are larger.

bin/build-runtime-library.sh
jwerle: this seems way too brittle. why can't we verify a successful build another way?

bin/publish-npm-modules.sh
jwerle: why can't we verify that the files we care about exist instead of counting how many there are?

bin/publish-npm-modules.sh
jwerle: I'm not sure on this one. Maybe @heapwolf or @chicoxyzzy has an idea about this. I've never done a check like this before

chicoxyzzy: I think that instead of comparing with the `PATH` version, we should do `ci_version_check.{sh|ps1}` sanity check on CI when we will have a build+publish action there.

chicoxyzzy: (we already do this check after we build Socket)

mribbons: > I think that instead of comparing with the `PATH` version, we should do `ci_version_check.{sh|ps1}` sanity check on CI when we will have a build+publish action there.

That was my intention here @chicoxyzzy.
Think about a situation where some `.js` or `.cc` outside `ssc` changes, code is committed, but the binary doesn't change.
If we then run `publish-npm-packages`, the shipped `ssc` won't match the latest commit.
It would be less of a problem in a CI generated build because everything would be built from source in the one off CI environment.

bin/build-runtime-library.sh
mribbons: Resolved as discussed.

bin/install.sh
mribbons: Resolved as discussed.

bin/publish-npm-modules.sh
mribbons: Resolved as discussed.

bin/publish-npm-modules.sh
mribbons: This one is still up for discussion. I have updated to test the package version rather than $PATH.

bin/publish-npm-modules.sh
```diff
@@ -63,13 +79,41 @@ mkdir -p "$SOCKET_HOME"
...
+  
+  if [[ ! -f "$lib_path/libsocket-runtime.a" ]]; then
+    ABORT_ERRORS=1
+    echo "$lib_path/libsocket-runtime.a missing - check build process."
+  fi
+  export ABORT_ERRORS
+done
+
+if (( ABORT_ERRORS )); then
+  echo "Refusing to publish due to missing static libs."
```
jwerle, line 115: I think

```sh
echo "Refusing to publish due to errors."
```

makes more sense because it looks like other things could cause `ABORT_ERRORS` to be `1`

bin/build-runtime-library.sh
```diff
@@ -100,7 +96,16 @@ while (( $# > 0 )); do
...
-if [[ "$host" = "Darwin" ]]; then
+if [[ "$platform" = "android" ]]; then
+  source "$root/bin/android-functions.sh"
+
+  if [[ ! -z $DEPS_ERROR ]]; then
+    echo "Android dependencies not satisfied."
+    exit 1
+  fi
+
+  clang=$(android_clang $ANDROID_HOME $NDK_VERSION $host $host_arch $arch "++")
```
jwerle, line 107:
```sh
clang="$(android_clang $ANDROID_HOME $NDK_VERSION $host $host_arch $arch "++")"
```

bin/build-runtime-library.sh
```diff
@@ -196,11 +173,14 @@ function main () {
...
 
-  declare static_library="$root/build/$arch-$platform/lib$d/libsocket-runtime$d.a"
+  declare base_lib="libsocket-runtime";
+  declare static_library="$root/build/$arch-$platform/lib$d/$base_lib$d.a"
   mkdir -p "$(dirname "$static_library")"
   declare ar="ar"
 
-  if [[ "$host" = "Win32" ]]; then
+  if [[ "$platform" = "android" ]]; then
+    ar=$(android_ar $ANDROID_HOME $NDK_VERSION $host $host_arch)
```
jwerle, line 182:
```sh
    ar="$(android_ar $ANDROID_HOME $NDK_VERSION $host $host_arch)"
```

bin/install.sh
```diff
@@ -2,10 +2,22 @@
...
+
+if [[ ! -z $VERBOSE ]]; then
+  echo "Using cores: $CPU_CORES"
+fi
+
 declare args=()
 declare pids=()
 declare force=0
 declare pass_force=""
+declare host=$(host_os)
```
jwerle, line 20:
```sh
    ar="$(android_ar $ANDROID_HOME $NDK_VERSION $host $host_arch)"
```

bin/install.sh
```diff
@@ -2,10 +2,22 @@
...
 
+source "$root/bin/functions.sh"
+source "$root/bin/android-functions.sh"
+
+if [[ -z "$CPU_CORES" ]]; then
+  export CPU_CORES=$(set_cpu_cores)
+fi
+
+if [[ ! -z $VERBOSE ]]; then
+  echo "Using cores: $CPU_CORES"
```
jwerle, line 13: Can we match the output style:

```sh
  echo "# using cores: $CPU_CORES"
```

bin/install.sh
```diff
@@ -169,7 +135,7 @@ function advice {
   elif [[ "$(uname -s)" == *"Linux"* ]]; then
     echo "apt-get install $1"
   elif [[ "$(uname -s)" == *"_NT"* ]]; then
-    echo "choco install $1 ?"
+    echo "Use install.ps1 to install $1."
```
jwerle, line 138: Maybe advise the command similarly to the above variants?

bin/install.sh
```diff
@@ -194,6 +160,20 @@ if [ "$host" == "Linux" ]; then
   die $? "not ok - missing pkg-config tool, \"$(advice 'pkg-config')\""
 fi
 
+if [[ -n "$BUILD_ANDROID" ]]; then
+  abis=($(android_supported_abis))
+  platform="android"
+  arch="${abis[0]}"
+  host_arch=$(uname -m)
+  clang=$(android_clang $ANDROID_HOME $NDK_VERSION $host $host_arch $arch)
```
jwerle, line 168:
```sh
  clang="$(android_clang $ANDROID_HOME $NDK_VERSION $host $host_arch $arch)"
```

jwerle (changes requested):
Left some comments. Generally, LGTM! However, we should quote pretty much everything in the case there is whitespace in an argument value, like a path causing the arguments to split.

bin/install.sh
mribbons: @jwerle, how about:

I'm pretty sure this is a noop, I just wanted to remove the choco reference.

I'll remove it entirely.

bin/build-runtime-library.sh
```diff
@@ -100,7 +96,16 @@ while (( $# > 0 )); do
   args+=("$arg")
 done
 
-if [[ "$host" = "Darwin" ]]; then
+if [[ "$platform" = "android" ]]; then
+  source "$root/bin/android-functions.sh"
+
+  if [[ -n $DEPS_ERROR ]]; then
+    echo "Android dependencies not satisfied."
```
jwerle, line 103:
```sh
    echo >&2 "not ok - Android dependencies not satisfied."
```

bin/install.sh
```diff
@@ -484,26 +477,42 @@ function _install {
...
+    fi
+    if [[ $host=="Win32" ]] && [[ $platform == "desktop" ]]; then
+      cp -rfp "$BUILD_DIR/$arch-$platform"/lib$_d/*.lib "$SOCKET_HOME/lib$_d/$arch-$platform"
     fi
+
+    if [[ $platform == "android" ]] && [[ -d "$BUILD_DIR/$arch-$platform"/lib ]]; then
+      cp -fr "$BUILD_DIR/$arch-$platform"/lib/*.a "$SOCKET_HOME/lib/$arch-$platform"
+    fi
+  else 
+    echo "no $BUILD_DIR/$arch-$platform/lib"
```
jwerle, line 502:
```sh
    echo >&2 "not ok - Missing $BUILD_DIR/$arch-$platform/lib"
```

bin/install.sh
```diff
@@ -565,25 +574,102 @@ function _setSDKVersion {
...
+    else
+      echo "failed to build $static_library"
+      exit 1
+    fi
+  
+  else
+    if [ -f $static_library ]; then
+      echo "ok - using cached static library ($arch-$platform): $(basename "$static_library")"
+    else
+      echo "static library doesn't exist after cache check passed: ($arch-$platform): $(basename "$static_library")"
```
jwerle, line 648:
```sh
      echo >&2 "not ok - static library doesn't exist after cache check passed: ($arch-$platform): $(basename "$static_library")"
```

bin/install.sh
```diff
@@ -565,25 +574,102 @@ function _setSDKVersion {
...
+    build_static=1
+  fi
+  mkdir -p "$(dirname $static_library)"
+
+  if (( build_static )); then
+    quiet $ar crs "$static_library" "${objects[@]}"
+    if [ -f $static_library ]; then
+      echo "ok - built $base_lib ($arch-$platform): $(basename "$static_library")"
+    else
+      echo "failed to build $static_library"
```
jwerle, line 640:
```sh
      echo >&2 "not ok - failed to build $static_library"
```

bin/publish-npm-modules.sh
```diff
@@ -63,13 +79,41 @@ mkdir -p "$SOCKET_HOME"
...
+declare ABORT_ERRORS=0
+
+
+# Confirm that build SSC matches current commit
+if command -v ssc >/dev/null 2>&1; then
+  REPO_VERSION="$(cat "$root/VERSION.txt") ($(git rev-parse --short=8 HEAD))"
+  BUILD_VERSION=$("$SOCKET_HOME"/bin/ssc --version)
+
+  if [[ "$REPO_VERSION" != "$BUILD_VERSION" ]]; then
+    echo "Repo $REPO_VERSION and $SOCKET_HOME/bin/ssc $BUILD_VERSION don't match."
```
jwerle, line 95:
```sh
    echo >&2 "not ok - Repository $REPO_VERSION and $SOCKET_HOME/bin/ssc $BUILD_VERSION don't match."
```

bin/publish-npm-modules.sh
```diff
@@ -63,13 +79,41 @@ mkdir -p "$SOCKET_HOME"
...
+    ABORT_ERRORS=1
+  fi
+fi
+
+for abi in $(android_supported_abis); do
+  lib_path="$SOCKET_HOME/lib/$abi-android"
+  if [[ ! -f "$lib_path/libuv.a" ]]; then
+    ABORT_ERRORS=1
+    echo "$lib_path/libuv.a missing - check build process."
```
jwerle, line 104:
```sh
    echo >&2 "not ok - $lib_path/libuv.a missing - check build process."
```

bin/publish-npm-modules.sh
```diff
@@ -63,13 +79,41 @@ mkdir -p "$SOCKET_HOME"
...
+  fi
+  
+  if [[ ! -f "$lib_path/libsocket-runtime.a" ]]; then
+    ABORT_ERRORS=1
+    echo "$lib_path/libsocket-runtime.a missing - check build process."
```
jwerle, line 109:
```sh
    echo >&2 "not ok - $lib_path/libsocket-runtime.a missing - check build process."
```

bin/publish-npm-modules.sh
```diff
@@ -63,13 +79,41 @@ mkdir -p "$SOCKET_HOME"
...
+  fi
+  export ABORT_ERRORS
+done
+
+if (( ABORT_ERRORS )); then
+  echo "Refusing to publish due to errors."
```
jwerle, line 115:
```sh
  echo >&2 "not ok - Refusing to publish due to errors."
```

src/cli/cli.cc
```diff
@@ -1363,15 +1418,24 @@ int main (const int argc, const char* argv[]) {
       }
     }
 
-    if (flagRunUserBuildOnly == false && fs::exists(paths.platformSpecificOutputPath)) {
-      auto p = fs::current_path() / fs::path(paths.platformSpecificOutputPath);
+    auto removePath = paths.platformSpecificOutputPath;
+      if (targetPlatform == "android")
+        removePath = settings["output"] + "/android/app/src";
```
jwerle, line 1423: is this path with `/` slashes okay for windows?

jwerle (commented):
Just a few cleanups and ready to merge!

src/cli/cli.cc
mribbons: Yes, it just works ;)

PR #235, opened 2023-03-12
Path implementation

jwerle:


test/src/path.js
```diff
@@ -11,5 +10,246 @@ test('path', (t) => {
...
+  )
+
+  t.equal(
+    path.win32.normalize('c:\\beep\\boop\\a\\b\\..\\..\\foo\\bar.js'),
+    'c:\\beep\\boop\\foo\\bar.js',
+    'relative Windows directory'
+  )
+
+  t.equal(
+    path.win32.normalize('c:\\beep\\boop\\a\\b\\..\\..\\foo\\bar\\'),
```
mribbons, line 237: I would say preserving the trailing slash is not required.
Including the trailing slash is a pre `path.join()` anti pattern.
We can make a decision here and say that a normalized path is one without a trailing slash.

test/src/process.js
```diff
@@ -16,9 +16,6 @@ test('process.exit()', (t) => {
 })
 
 test('process.cwd', async (t) => {
```
mribbons, line 18: Can  we add a test to ensure `cwd` exists?

test/src/fs/promises.js
```diff
@@ -65,8 +65,9 @@ if (process.platform !== 'ios') {
 
   test('fs.promises.readFile', async (t) => {
     const data = await fs.readFile(FIXTURES + 'file.txt')
+    const crlf = process.platform === 'win32' ? '\r\n' : '\n'
```
mribbons, line 68: `crlf` should be `eol` :)

test/src/path.js
jwerle: We want to maintain parity with Node's `path` module here to keep up expectations. I also would think that if I give a trailing slash, that I get one back.

test/src/process.js
jwerle: there are plenty tests for this already in `test/src/process.js`

test/src/fs/promises.js
jwerle: 👍🏼

jwerle:
Okay this one is good to go. Tests passing for all platforms!

mribbons (approved):

chicoxyzzy (approved):
:tada:

PR #237, opened 2023-03-13
bin: cleanup

mribbons:
bin: convert some bash scripts to functions

call from the command line using ./bin/functions --old-script-name

This is an attempt to cut down the clutter in `bin` as per @heapwolf


bin/functions.sh
```diff
@@ -0,0 +1,129 @@
...
+    exit 0;
+  else
+    echo "Version check has failed";
+    echo "Expected: $VERSION_EXPECTED";
+    echo "Got: $VERSION_SSC";
+    exit 1;
+  fi
+}
+
+function generate_api_import_map() {
```
chicoxyzzy, line 21: I think we don't use import maps anymore

heapwolf:
awesome. @jwerle thoughts?

jwerle:
> awesome. @jwerle thoughts?

maybe we can call `function.sh` something else? something smaller? cc @chicoxyzzy

mribbons:
> > awesome. @jwerle thoughts?
> 
> maybe we can call `function.sh` something else? something smaller? cc @chicoxyzzy

Just FYI, this will be combined with `functions.sh` from https://github.com/socketsupply/socket/pull/234, but no problem to rename.

mribbons:
> I'm not really sold on the idea of moving everything to a single file. It is a terrible experience to scroll through a single file that has multiple concerns, much like cli.cc.

Fair point, Maybe what we need is a bit of documentation in each script.

> Also, I noticed an important script in my workflow was deleted, and not migrated to `functions.sh`

Good save @jwerle!

Let's cancel this PR? I don't want to manually add anything back as it will lose the git history.

I can redo the powershell stuff using  `cherry-pick`

mribbons:
> I'm not really sold on the idea of moving everything to a single file. It is a terrible experience to scroll through a single file that has multiple concerns, much like `cli.cc`.

I think another drawback to this approach is that we would lose tab completion support on these util scripts, also not great for dev xp.

jwerle:
> > I'm not really sold on the idea of moving everything to a single file. It is a terrible experience to scroll through a single file that has multiple concerns, much like cli.cc.
> 
> Fair point, Maybe what we need is a bit of documentation in each script.
> 
> > Also, I noticed an important script in my workflow was deleted, and not migrated to `functions.sh`
> 
> Good save @jwerle!
> 
> Let's cancel this PR? I don't want to manually add anything back as it will lose the git history.
> 
> I can redo the powershell stuff using `cherry-pick`

yeah that sounds good to me!

jwerle:
> > I'm not really sold on the idea of moving everything to a single file. It is a terrible experience to scroll through a single file that has multiple concerns, much like `cli.cc`.
> 
> I think another drawback to this approach is that we would lose tab completion support on these util scripts, also not great for dev xp.

yeah we'd need to implement a complete function 😓

jwerle:
> 

closing as per @mribbons

jwerle:
@mribbons feel free to delete this branch when you are ready

PR #239, opened 2023-03-14
Add --config option to the ssc init

chicoxyzzy:
This helps to wrap the existing application into Socket Runtime

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -960,7 +960,7 @@ int main (const int argc, const char* argv[]) {
         }
 
         // default values
-        settings["build_output"] = settings["build_output"].size() > 0 ? settings["build_output"] : "dist";
+        settings["build_output"] = settings["build_output"].size() > 0 ? settings["build_output"] : "build";
         settings["meta_lang"] = settings["meta_lang"].size() > 0 ? settings["meta_lang"] : "en-us";
         settings["meta_version"] = settings["meta_version"].size() > 0 ? settings["meta_version"] : "1.0.0";
         settings["meta_title"] = settings["meta_title"].size() > 0 ? settings["meta_title"] : settings["build_name"];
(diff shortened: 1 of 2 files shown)


jwerle:
Maybe it's worthwhile just checking if files exist before over writing them?

getify:
I agree with @jwerle here. I think the creation of the directory and files is helpful but should be guarded on whether they don't yet exist, rather than either a potential error OR (worse) overwriting them.

src/cli/cli.cc
```diff
@@ -960,7 +960,7 @@ int main (const int argc, const char* argv[]) {
         }
 
         // default values
-        settings["build_output"] = settings["build_output"].size() > 0 ? settings["build_output"] : "dist";
+        settings["build_output"] = settings["build_output"].size() > 0 ? settings["build_output"] : "build";
```
chicoxyzzy, line 963: so now we have `"build"` by default in both `.gitignore` and as an implicit default value

chicoxyzzy:
made it exit early when `socket.ini` exists and skip creating `.gitignore` and `src/index.html` if they exist

getify (approved):
LGTM

jwerle (approved):

src/cli/templates.hh
```diff
@@ -1379,14 +1382,15 @@ flags = -O3
...
+; A directory where your application's code is located. ssc will copy everything in this directory
+; to the build output directory. It's ignored if the "script" property is set.
 input = "src"
 
 ; The name of the program and executable to be output. Can't contain spaces or special characters. Required field.
 name = "beepboop"
 
 ; The binary output path. It's recommended to add this path to .gitignore.
-output = "dist"
+output = "build"
```
jwerle, line 1393: finally, thank you

jwerle (approved):

PR #242, opened 2023-03-15
remove obsolete property from default config

chicoxyzzy:


chicoxyzzy:
going to just merge it

YOLO :D

PR #244, opened 2023-03-16
Initial build copy map, drop dep on node for tests

jwerle:
In an effort to drop dependency on Node, this pull request introduces a feature called _Build Copy Map_ that allows the specification of a mapping of input files to relative output files in the resource build directory. This replaces the usage of `esbuild` in our test suite and hopefully for other simple copy use cases downstream.

A build copy_map file is INI and can be specified in the `[build] copy_map` configuration. The path is relative to the root of the project (`targetPath`) and looks like:

_test/src/mapping.ini_:

```ini
## for the copy-map tests
./copy-map/file-a.js = "copy-map/A.js"
./copy-map/file-b.js = "copy-map/B.js"

## for configuration tests
../socket.ini = "."

## test fixtures
../fixtures = "."
```

which is now used in the test suite, removing usage of `esbuild` which did this job.


mribbons:
Windows, the tests don't run, there's no error output in terminal.
I get these errors in js console:
```
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "". Strict MIME type checking is enforced for module scripts per HTML spec. ﻿
diagnostics/:1 

Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "". Strict MIME type checking is enforced for module scripts per HTML spec.
fs/:1 
```

mribbons (changes requested):
Windows, the tests don't run, there's no error output in terminal.
I get these errors in js console:
```
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "". Strict MIME type checking is enforced for module scripts per HTML spec. ﻿
diagnostics/:1 

Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "". Strict MIME type checking is enforced for module scripts per HTML spec.
fs/:1 
```

src/cli/cli.cc
```diff
@@ -2172,6 +2197,61 @@ int main (const int argc, const char* argv[]) {
...
+        log("warning: file specified in [build] copy_map does not exist");
+      } else {
+        auto copyMap = trim(readFile(copyMapFile));
+        auto copyMapFileDirectory = fs::absolute(copyMapFile.relative_path().parent_path());
+
+        if (copyMap.size() > 0) {
+          auto lines = split(copyMap, '\n');
+          for (const auto& line : lines) {
+            // handle :|#|// leading comments
+            if (line.starts_with(":") || line.starts_with("#") || line.starts_with("//")) {
```
mribbons, line 2212: Is there a reason for not adding this functionality to `parseConfig()`

If so, how about implementing this parser functionality in another function so we can use it for `.sscrc` as per discord?

src/cli/cli.cc
```diff
@@ -2152,16 +2152,41 @@ int main (const int argc, const char* argv[]) {
 
     if (settings.count("build_copy") != 0) {
       fs::path pathInput = settings["build_copy"].size() > 0
-        ? targetPath / settings["build_copy"]
-        : targetPath / "src";
```
chicoxyzzy, line 2156: was `targetPath` unnecessary here?

src/cli/cli.cc
```diff
@@ -2152,16 +2152,41 @@ int main (const int argc, const char* argv[]) {
...
+        }
+
+        fs::copy(
+          src,
+          dst,
+          fs::copy_options::update_existing | fs::copy_options::recursive
+        );
+      }
     // @deprecated
-    } else if (settings.count("build_input") != 0 && settings.count("build_script") == 0) {
```
chicoxyzzy, line 2163: it was necessary for backward compatibility. ssc was copying files from `input` only when `script` is not presented

chicoxyzzy: actually I think we can remove this `esle if` completely. The feature `input` copy was never documented

test/src/mapping.ini
chicoxyzzy, line 1: does it work without quotes as well?

jwerle: yep!

src/cli/cli.cc
jwerle: well it is INI without sections, more just `key = value`

test/socket.ini
```diff
@@ -1,7 +1,7 @@
 [build]
 name = "socket-runtime-javascript-tests"
-input = src
-script = node build.js
+copy = src
+copy_map = src/mapping.ini
```
chicoxyzzy, line 4: what happens when both are presented?

jwerle: they both do what you'd expect

chicoxyzzy: are you copying `src/mapping.ini` to the output dir for testing purposes?

src/cli/cli.cc
jwerle: Going to leave as is as the test suite depended on this before the deprecation

src/cli/cli.cc
mribbons: Even if you aren't using sections, wouldn't the two feature sets (`parseConfig` and what you have implemented) still be compatible and have a fair bit of overlap?
It looks as though `parseConfig` has support for empty prefixes (global section)
Also you didn't answer the query about at least implementing a separate function that we can use for a `.sscrc` file as well as this mapping config.

mribbons (approved):
Wow, windows CI passing! 👏 

I really think you ought to consider this one a bit more in terms of technical debt:
https://github.com/socketsupply/socket/pull/244#discussion_r1141600938

chicoxyzzy (approved):

src/cli/cli.cc
chicoxyzzy: what is `.sscrc`?

PR #260, opened 2023-03-21
use the same style for warnings and errors in cli

chicoxyzzy:
also use one style for all the warnings and error (all capital followed by colon)

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -189,7 +189,7 @@ String getAndroidHome() {
     setenv("ANDROID_HOME", androidHome.c_str(), 1);
     #endif
 
-    log("warning: 'ANDROID_HOME' is set to '" + androidHome + "'");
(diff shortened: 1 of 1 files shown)


jwerle (approved):

src/cli/cli.cc
```diff
@@ -333,7 +333,7 @@ int runApp (const fs::path& path, const String& args, bool headless) {
         appMutex.unlock();
         appStatus = 1;
         debug(
-          "error: NSWorkspace: (code=%lu, domain=%@) %@",
+          "ERROR: NSWorkspace: (code=%lu, domain=%@) %@",
```
chicoxyzzy, line 336: is this ok to apply this change for `debug` as well?

jwerle: yeah it shouldn't break anything

PR #262, opened 2023-03-21
Read '.ssc.env' if present

jwerle:
This PR introduces support for reading environment variables, potentially determined in external build pipelines (like npm). If a `.ssc.env` file is present in the working target directory, it is read and the environment variables are added to the execution environment of the `ssc` cli. An example file may look like:

```ini
JAVA_HOME="..."
ANDROID_HOME="..."
```

This file differs from the `[env]` section of `.sscrc` as this file is most likely programmatically generated where `.sscrc` is user controlled

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -787,9 +787,8 @@ inline String getCfgUtilPath() {
 }
 
 void initializeEnv (fs::path targetPath) {
-  auto SSC_ENV_FILENAME = getEnv("SSC_ENV_FILENAME");
-
-  auto filename = SSC_ENV_FILENAME.size() > 0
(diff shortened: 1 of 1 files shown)


mribbons (approved):
Thanks @jwerle!

src/cli/cli.cc
```diff
@@ -786,13 +786,45 @@ inline String getCfgUtilPath() {
   exit(1);
 }
 
+void initializeEnv (fs::path targetPath) {
+  static bool initialized = false;
+
+  if (initialized) return;
```
jwerle, line 792: @mribbons we should probably remove this check now!

chicoxyzzy (approved):

PR #264, opened 2023-03-21
Improve preload injection

jwerle:
This PR ensures the preload is evaluated in all possible loaded frames, and before _any_ JavaScript is executed. This PR also patches `queueMicrotask` to ensure any error thrown in the callback is properly bubbled to the global `globalThis` object. cc @chicoxyzzy


api/internal/init.js
```diff
@@ -1,93 +1,65 @@
...
+          return callback.call(globalThis)
+        } catch (error) {
+          // XXX(@jwerle): `queueMicrotask()` is broken in WebKit WebViews
+          // If an error is thrown, it does not bubble to the `globalThis`
+          // object, but is instead silently discarded. This is misleading
+          // and results in confusion for developers trying to debug something
+          // they may have done wrong. Here we will rethrow the exception out
+          // of microtask execution context with the best function we can
+          // possibly use to report the exception to the `globalThis` object
+          // as an `Unhandled Promise Rejection` error
```
chicoxyzzy, line 57: 👍 

I think we should report the bug on the https://bugs.webkit.org, but I'm not sure what's the minimal repro is

chicoxyzzy: oh, found it in the microtask tests

chicoxyzzy (approved):

chicoxyzzy:
I can see errors now 👍

PR #292, opened 2023-03-29
parse INI list values consistently

jwerle:
Lists taking the form of:

```ini
value = 1, 2, 3
```

or

```ini
value = 1 2 3
```

or

```ini
value[] = 1
value[] = 2
value[] = 3
```

These variants were not parsed correctly leading to unexpected behavior.


bin/update-stream-relay-source.sh
mribbons, line 1: Should these changes be on a separate commit?

mribbons (approved):

PR #293, opened 2023-03-30
fix race condition when loading node:crypto

missinglink:
heya,

I noticed this race when using `api/crypto.js` from node.

```js
import { getRandomValues } from './api/crypto.js'
console.error(getRandomValues(new Uint8Array(8)))
```

```js
Missing implementation for window.crypto.getRandomValues()
null
```

I think it's not triggered in the test suite because it executes on the next tick 🤔

### DIFF
--- a/api/crypto.js
+++ b/api/crypto.js
@@ -14,9 +14,8 @@ import * as exports from './crypto.js'
 let getRandomValuesFallback = null
 
 if (globalThis?.process?.versions?.node) {
-  import('node:crypto').then((c) => {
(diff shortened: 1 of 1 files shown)


jwerle:
hey @missinglink! Good to see you here. Nice catch. This works and we'll happily accept this PR as we don't care about async ESM in a Node context, but generally WRT a WebView context, we wanted to avoid top level await in an ESM module

jwerle (approved):

mribbons (approved):

chicoxyzzy (approved):

PR #298, opened 2023-04-04
Remove top-level-await usage, introduce 'crypto.ready' promise

jwerle:
It looks like #293 re-introduced the top level await issue we observed in the past causing ESM to finishing loading _after_ `window.onload` is called. This PR fixes that and introduces a `crypto.ready` promise that resolves when pending async operations are done.

```js
import crypto from 'socket:crypto'
await crypto.ready
```

cc @missinglink

### DIFF
--- a/api/crypto.js
+++ b/api/crypto.js
@@ -1,7 +1,7 @@
 /**
  * @module Crypto
  *
- * Some high-level methods around the `crypto.subtle`API for getting
+ * Some high-level methods around the `crypto.subtle` API for getting
  * random bytes and hashing.
  */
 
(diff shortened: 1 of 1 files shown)


chicoxyzzy (approved):

missinglink:
thanks for the mention, works for me 👍 

I'm curious if `await ready` is equivalent to `await new Promise(process.nextTick)` in this context.

Both methods allow me to wait for the import to be loaded, although I guess `await ready` is the preferred method as the latter can't be guaranteed to resolve within a single tick, or can it? 🤔

jwerle:
> thanks for the mention, works for me 👍
> 
> I'm curious if `await ready` is equivalent to `await new Promise(process.nextTick)` in this context.
> 
> Both methods allow me to wait for the import to be loaded, although I guess `await ready` is the preferred method as the latter can't guarantee the import resolves within a single tick, or can it? 🤔

Maybe in node, and possibly in the webview, but this should give us future room to include more async setup 🤞

PR #301, opened 2023-04-05
This PR completes FTE processes for all install paths

mribbons:


npm/bin/ssc.js
```diff
@@ -23,6 +23,11 @@ async function main () {
     ...process.env
   }
 
+  if (!await installation.firstTimeExperienceSetup()) {
+    // FTE not completed satisfactorally, or 'setup' was ran externally
```
chicoxyzzy, line 27: should we log something here?

npm/src/index.js
```diff
@@ -21,9 +23,66 @@ export const bin = {
...
+
+  // env doesn't exist, attempt to run setup for target being built. This should also install vc_redist if it hasn't been installed
+  const PLATFORM_PARAMETER = '--platform='
+  let isSetupCall = false
+
+  let platform = ''
+  process.argv.slice(2).forEach(arg => {
+    if (arg === 'setup') {
+      isSetupCall = true
+    } if (arg.indexOf(PLATFORM_PARAMETER) === 0) {
```
chicoxyzzy, line 40: nit: missing line break

npm/src/index.js
```diff
@@ -21,9 +23,66 @@ export const bin = {
     : path.join(env.PREFIX, 'bin', 'ssc')
 }
 
+export const firstTimeExperienceSetup = async () => {
```
chicoxyzzy, line 26: I feel like we can improve naming here to make it more clear what the return value mean

src/cli/cli.cc
```diff
@@ -1611,8 +1664,8 @@ int main (const int argc, const char* argv[]) {
     }
 
     auto removePath = paths.platformSpecificOutputPath;
-      if (targetPlatform == "android")
-        removePath = settings["output"] + "/android/app/src";
+    if (targetPlatform == "android")
+        removePath = settings["build_output"] + "/android/app/src";
```
chicoxyzzy, line 1668: good catch!

src/cli/cli.cc
```diff
@@ -2838,49 +2908,52 @@ int main (const int argc, const char* argv[]) {
...
+              << " APP_PLATFORM=" << androidPlatform
+              << " NDK_LIBS_OUT=" << jniLibs
+            ;
+
+            if (!(debugEnv || verboseEnv)) ndkBuildArgs << " >" << (!platform.win ? "/dev/null" : "NUL") << " 2>&1";
+
+            if (debugEnv || verboseEnv) log(ndkBuildArgs.str());
+            if (std::system(ndkBuildArgs.str().c_str()) != 0) {
+              log(ndkBuildArgs.str());
+              log("ndk build failed.");
```
chicoxyzzy, line 2954: nit: 
```suggestion
              log("ERROR: ndk build failed.");
```

src/cli/cli.cc
```diff
@@ -3627,6 +3700,86 @@ int main (const int argc, const char* argv[]) {
       exit(exitCode);
     }
   });
+  
+  createSubcommand("setup", { "--platform", "--yes" }, false, [&](const std::span<const char *>& options) -> void {
```
chicoxyzzy, line 3704: `"-y"` is missing in a vector of valid options. CLI will print an error that `"-y"` is not valid.

src/cli/cli.cc
```diff
@@ -3627,6 +3700,86 @@ int main (const int argc, const char* argv[]) {
...
+      targetAndroid = true;
+    } else if (is(targetPlatform, "windows")) {
+      targetWindows = true;
+    } else {
+      help = true;
+    }
+
+    if (help) {
+      printHelp("setup");
+      exit(1);
```
chicoxyzzy, line 3743: should be `exit(0)`

src/cli/cli.cc
```diff
@@ -3627,6 +3700,86 @@ int main (const int argc, const char* argv[]) {
...
+    }
+
+    if (!platform.win && targetWindows) {
+      std::cout << "Windows build dependencies can only be installed on Windows." << std::endl;
```
chicoxyzzy, line 3747: nit: you can use `log` here and maybe add `WARNING: ` or `ERROR: ` prefix

src/cli/cli.cc
```diff
@@ -3627,6 +3700,86 @@ int main (const int argc, const char* argv[]) {
...
+      if (targetAndroid) {
+        script = prefixFile("./bin/android-functions.sh");
+      }
+      yesArg = yes ? "--yes-deps" : "";
+    }
+
+    script = fs::path(script.string().substr(0, script.string().size()-1));
+
+    if (!fs::exists(script)) {
+      std::cout << "Install script not found: '" << script.string() << "'" << std::endl;
```
chicoxyzzy, line 3771: same here

src/cli/cli.cc
```diff
@@ -3627,6 +3700,86 @@ int main (const int argc, const char* argv[]) {
...
+      exit(1);
+    }
+
+    fs::current_path(prefixFile());
+
+    String command = scriptHost + " \"" + script.string() + "\" " + argument + " " + yesArg;
+    std::cout << command << std::endl;
```
chicoxyzzy, line 3778: and here

src/cli/templates.hh
```diff
@@ -117,6 +118,22 @@ options:
...
+Setup build tools for target <platform>
+
+Platforms not listed below can be setup using instructions at https://socketsupply.co/guides
+
+usage:
+  ssc setup --platform=<platform> [--yes|-y]
+
+options:
+  --platform       windows|android
+  --yes|-y         answer yes to any prompts
```
chicoxyzzy, line 133: nit: we use these format in other places in readme: `-y, --yes`

src/cli/templates.hh
```diff
@@ -117,6 +118,22 @@ options:
...
+
+Setup build tools for target <platform>
+
+Platforms not listed below can be setup using instructions at https://socketsupply.co/guides
+
+usage:
+  ssc setup --platform=<platform> [--yes|-y]
+
+options:
+  --platform       windows|android
```
chicoxyzzy, line 132: `--platform<=platform> target (windows, android)`

src/cli/templates.hh
chicoxyzzy, line 1: while you are here, can you please also fix this line in the `ssc build -h`:
`--platform<=platform>` => `--platform=<platform>`
🙏

src/cli/cli.cc
```diff
@@ -2838,49 +2908,52 @@ int main (const int argc, const char* argv[]) {
         auto libs = _main / "libs";
         auto obj = _main / "obj";
 
+        if (!flagRunUserBuildOnly || !fs::exists(jniLibs)) {
+
         if (fs::exists(obj)) {
```
chicoxyzzy, line 2913: nit: identation

npm/bin/ssc.js
mribbons: @chicoxyzzy The user will be able to see the results of `install.sh` or `install.ps1` at this time.

npm/src/index.js
mribbons: Hm should have been `else if`.

npm/src/index.js
mribbons: I'm open to suggestions. It was initially just an action method, now it's returning the status to indicate whether or not setup is complete enough to run `ssc`: On windows the min requirement is for `VC_Runtime140.dll` to be installed.

`firstTimeExperienceSetupIsValid`?

src/cli/cli.cc
mribbons: My thinking here is that if a `ssc setup` is scripted with the wrong arguments, we want to return an exit code indicating failure.

src/cli/cli.cc
mribbons: Great spot.

npm/src/index.js
chicoxyzzy: maybe something like `areSetupPrerequisitesMet`?

src/cli/cli.cc
chicoxyzzy: ah, ok, that makes sense

npm/src/index.js
mribbons: I'm not keen on something like `areSetupPrerequisitesMet` because it doesn't inform that FTE is going to be triggered as part of this function.

jwerle (approved):

npm/src/index.js
jwerle: `firstTimeExperienceSetup` sounds find to me

src/cli/cli.cc
```diff
@@ -3648,6 +3731,86 @@ int main (const int argc, const char* argv[]) {
...
+      if (platform.size() > 0) {
+        targetPlatform = platform;
+      }
+
+      if (is(arg, "-y") || is(arg, "--yes")) {
+        yes = true;
+      }
+
+      if (help) {
+        break;
```
chicoxyzzy, line 3758: in this case it should just exit with 0 IMHO
```cpp
     printHelp("setup");
     exit(0);
```

chicoxyzzy: maybe just after
```cpp
if (is(arg, "-h") || is(arg, "--help")) {
```

chicoxyzzy (approved):

src/cli/cli.cc
mribbons: @jwerle Can you tie break this one please?

mribbons: @chicoxyzzy Sorry I get it now. Fixed I think.

PR #305, opened 2023-04-05
Handle aarch64 -> arm64 host arch for Android builds, add shell proxy test runner

jwerle:


bin/android-functions.sh
```diff
@@ -46,7 +46,7 @@ function android_arch() {
     x86)
       echo -n "i686"
       ;;
-    arm64|x86-64)
+    aarch64|arm64|x86-64)
```
mribbons, line 49: Could you put a comment here:
```diff
+    # linux / aarch64 not currently supported: https://github.com/socketsupply/socket/issues/306
+    aarch64|arm64|x86-64)
```

mribbons (changes requested):

bin/android-functions.sh
mribbons: @jwerle actually I believe this change isn't required, because host_arch() should have already converted `aarch64` to `arm64`:

https://github.com/socketsupply/socket/blob/handle-aarch64-android/bin/functions.sh#L92
```bash
function host_arch() {
  uname -m | sed 's/aarch64/arm64/g'
}
```

https://github.com/socketsupply/socket/blob/handle-aarch64-android/bin/android-functions.sh#L87
`android_clang` is called with `$(host_arch)` in position 4, which is used to build the compiler path.
```bash
echo "$ANDROID_HOME/ndk/$NDK_VERSION/toolchains/llvm/prebuilt/$(android_host_platform "$host")-$(android_arch "$host_arch")/bin/clang$plusplus --target=$(android_arch "$arch")-linux-android$(android_eabi "$arch")"
```

https://github.com/socketsupply/socket/blob/handle-aarch64-android/bin/install.sh#L605-L606
```bash
  local host_arch="$(host_arch)"
  clang="$(android_clang "$ANDROID_HOME" "$NDK_VERSION" "$host" "$host_arch" "$arch")"
```

mribbons (changes requested):

chicoxyzzy:
we probably can remove `armeabi` and `x86` Android archs builds

mribbons:
> we probably can remove `armeabi` and `x86` Android archs builds

If so we should make it configurable in `.ssc.env` or `.ssc.rc` and consider setting useful defaults.
`x86` is good for emulation on intel CPUs.

chicoxyzzy:
> x86 is good for emulation on intel CPUs.

we don't support `x86` platforms for `ssc`, only `x86_64` which is also presented in the build

chicoxyzzy (approved):

PR #307, opened 2023-04-06
install.sh: locate latest clang using dpkg

mribbons:


bin/install.sh
```diff
@@ -101,8 +101,26 @@ if [[ "$host" == "Win32" ]]; then
...
+    echo "sudo apt-get install $1"
+  fi
+}
+
 if [ ! "$CXX" ]; then
-  if command -v clang++ >/dev/null 2>&1; then
+  # TODO(@mribbons): yum support
+  if command -v dpkg >/dev/null 2>&1; then
+    # Convert clang++ paths to a path + version (bin_version), sort by version, then cut out bin path back out to get the highest installed clang version
+    CXX=$(dpkg -S clang 2>&1| grep "++" | cut -d" " -f 2 | while read clang; do bin_version="$("$clang" --version|head -n1)#$clang"; echo $bin_version; done | sort -r | cut -d"#" -f 2 | head -n1)
```
heapwolf, line 118: could this be splt into a few parts so its easier to read?

heapwolf (approved):
LGTM

bin/install.sh
```diff
@@ -101,8 +101,26 @@ if [[ "$host" == "Win32" ]]; then
...
 
+function advice {
+  if [[ "$(uname -s)" == "Darwin" ]]; then
+    echo "brew install $1"
+  elif [[ "$(uname -r)" == *"ARCH"* ]]; then
+    echo "sudo pacman -S $1"
+  elif [[ "$(uname -s)" == *"Linux"* ]]; then
+    echo "sudo apt-get install $1"
+  fi
+}
```
chicoxyzzy, line 112: we can add another else branch with something like `echo "to install $1 with your package manager"`

chicoxyzzy (approved):

jwerle (approved):

bin/install.sh
mribbons: @heapwolf trying to find a way to split it out, but CXX isn't persisted to the parent block, it always outputs as empty in this code:
```bash
    {
      dpkg -S clang 2>&1| grep "++" | cut -d" " -f 2 | while read clang; do
        bin_version="$("$clang" --version|head -n1)#$clang"
        echo $bin_version;
      done
    } | sort -r | cut -d"#" -f 2 | head -n1 | while read CXX; do echo "CXX: $CXX"; export CXX; done
    echo "CXX: $CXX"
 ```

Any ideas?

jwerle (approved):

PR #308, opened 2023-04-06
ssc-setup fixes

jwerle:
This PR improves `ssc setup` in various ways including the output during the FTE and setup to give users more feedback about what is going on aligning the output style with the rest of build scripts.

### DIFF
--- a/bin/android-functions.sh
+++ b/bin/android-functions.sh
@@ -438,14 +438,14 @@ function android_install_sdk_manager() {
     return 0
   fi
 
-  if ! prompt_yn "> The Android SDK Manager is required for building Android apps. Install it now?"; then
+  if ! prompt_yn "The Android SDK Manager is required for building Android apps. Install it now?"; then
     return 1
   fi
 
   write_log "h" "# Please review the Android SDK Manager License by visiting the URL below and clicking "Download SDK Platform-Tools for "[Your OS]"
   write_log "h" "# $ANDROID_PLATFORM_TOOLS_PAGE_URI"
 
-  if ! prompt_yn "> Do you consent to the Android SDK Manager License?"; then
+  if ! prompt_yn "Do you consent to the Android SDK Manager License?"; then
     return 1
   fi
 
@@ -454,7 +454,7 @@ function android_install_sdk_manager() {
     write_log "h" "# Using existing ANDROID_HOME: $ANDROID_HOME folder"
     _ah="$ANDROID_HOME"
   else
-    prompt_new_path "> Enter location for ANDROID_HOME" "${ANDROID_HOME_SEARCH_PATHS[0]}" _ah \
+    prompt_new_path "Enter location for ANDROID_HOME" "${ANDROID_HOME_SEARCH_PATHS[0]}" _ah \
       "If you want to use an existing path as ANDROID_HOME, set it in an environment variable before running this script."
 
     if [ -d "$_ah" ]; then
@@ -480,7 +480,7 @@ function android_install_sdk_manager() {
   fi
   rm -f "$archive"
 
-  if ! prompt_yn "> Android Command line tools is required for building Android apps. Install it now?"; then
+  if ! prompt_yn "Android Command line tools is required for building Android apps. Install it now?"; then
     return 1
   fi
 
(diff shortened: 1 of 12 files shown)


test/scripts/test-android-emulator.sh
```diff
@@ -2,8 +2,23 @@
...
+
+if [[ -n $1 ]]; then
+  host=$1
+else
+  host="$(uname -s)"
+fi
+
+if [[ "$host" = "Linux" ]]; then
+  if [ -n "$WSL_DISTRO_NAME" ] || uname -r | grep 'Microsoft'; then
+    echo >&2 "erroR: WSL is not supported."
```
chicoxyzzy, line 14:
```suggestion
    echo >&2 "ERROR: WSL is not supported."
```

bin/functions.sh
```diff
@@ -387,8 +387,8 @@ function lower() {
   echo "$1"|tr '[:upper:]' '[:lower:]'
 }
 
-function version { 
-  echo "$@" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }'; 
+function version() {
+  echo "$@" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }';
 }
```
chicoxyzzy, line 392: just curious, version of what?

chicoxyzzy (approved):
We need to document all the flags like `--only-env`, etc. It's hard to keep all this stuff in a head

jwerle:
> We need to document all the flags like `--only-env`, etc. It's hard to keep all this stuff in a head

we'll come back and do a complete pass on docs for the scripts 🙏🏼

bin/android-functions.sh
```diff
@@ -97,7 +100,7 @@ function test_javac_version() {
   fi
 
   jc_v="${jc_v/javac\ /}"
-  write_log "v" "Comparing $javac version: "$(version "$jc_v") "$(version "$target_version")"
+  write_log "v" "# Comparing $javac version: "$(version "$jc_v") "$(version "$target_version")"
   if [ "$(version "$jc_v")" -lt "$(version "$target_version")" ]; then
```
jwerle, line 104: @chicoxyzzy

bin/functions.sh
```diff
@@ -271,19 +271,19 @@ function prompt_new_path() {
...
-      write_log "h" "\"$input\" already exists, please choose a new path."
+      write_log "h" "# \"$input\" already exists, please choose a new path."
       if [ -n "$exists_message" ]; then
-        write_log "h" "$exists_message"
+        write_log "h" "# $exists_message"
       fi
       input=""
     elif [ -z "$input" ]; then
-      if prompt_yn "Cancel entering new path?"; then
+      if prompt_yn "> Cancel entering new path?"; then
```
mribbons, line 280: prompt_yn calls prompt which outputs >, so this will prompt with >>

bin/functions.sh
```diff
@@ -251,14 +251,14 @@ function prompt_new_path() {
...
-    if prompt_yn "Use suggested default path: ""$2""?"; then
+    if prompt_yn "> Use suggested default path: ""$2""?"; then
       input="$2"
     fi
   fi
 
   while true; do
     if [ -z "$input" ]; then
-      prompt "$text (Press Enter to go back)" input
+      prompt "> $text (Press Enter to go back)" input
```
mribbons, line 261: prompt outputs >, so this will prompt with >>

src/cli/cli.cc
```diff
@@ -114,6 +114,10 @@ String getSocketHome (bool verbose) {
   static String LOCALAPPDATA = getEnv("LOCALAPPDATA");
   static String SOCKET_HOME = getEnv("SOCKET_HOME");
   static String HOME = getEnv("HOME");
+  static const bool SSC_DEBUG = (
+    getEnv("SSC_DEBUG").size() > 0 ||
```
mribbons, line 118: Nice. I think we also need to look at differentiating between debug output and debug builds in `install.sh` and `install.ps1`.

src/cli/cli.cc
```diff
@@ -150,7 +164,11 @@ String getSocketHome () {
 }
 
 String getAndroidHome() {
-  auto androidHome = getEnv("ANDROID_HOME");
+  static auto androidHome = getEnv("ANDROID_HOME");
```
mribbons, line 167: Using `static` here will make unit testing / mocking harder in future.

mribbons (changes requested):

src/cli/cli.cc
jwerle: yeah I know what you mean, but let's punt on that for now until we can rethink this file

mribbons (approved):

PR #309, opened 2023-04-07
Support CommonJS and a subset of the stdlib as builtins

jwerle:
**Implementation:**

- relative path resolution (with or without extensions):
  - `./module`
  - `./module.js`
  - `./module/index.js`
  - `./module.json`
  - `./module/index.json`
- named modules
  - `require('foo')`
    - resolves to `file://foo` relative to `window.location.href` (following relative path resolution above)
    - also follows tries prefixes in `SOCKET_MODULE_PATH_PREFIX` (delimited with `:`)
      - if `SOCKET_MODULE_PATH_PREFIX="node_modules:bower_components"` then `require('foo')` will try `file://node_modules/foo` and `file://bower_components/foo` relative to `window.location.href` (following relative path resolution above) 
- loading builtins (limited)
  - `require('fs')` or `require('socket:fs')` (node prefix compat too eg: `node:fs`) returns `fs` ESM module default export, this is true for all builtins (see below)

**Builtins:**

Preloaded builtins can be seen below. They cannot be overloaded

```js
export const builtins = {
  buffer,
  console,
  dgram,
  dns,
  events,
  fs,
  ipc,
  module,
  os,
  path,
  process,
  stream,
  test,
  util
}
```

**Extra features:*

- top level await in module
- modules are `EventTarget` instances so `load` and `error` events fire on them (on `module` not `module.exports`)
- error bubbling `child module -> parent module -> ... Module.main -> globalThis` 
- custom resolvers: `Module.resolvers.push((specifier, ctx, next) => { ... })`

```js
// this just works
const fs = require('fs')
fs.readdir(process.cwd(), console.log)
```


chicoxyzzy (approved):
I wish we have ShadowRealms in browsers sooner

PR #327, opened 2023-04-10
Fix CommonJS module path normalization

jwerle:


api/module.js
```diff
@@ -33,10 +33,18 @@ import util from './util.js'
  * @typedef {(specifier: string, ctx: Module, next: function} => undefined} ModuleResolver
  */
 
+function normalizePathname (pathname) {
+  if (os.platform()) {
+    return path.win32.normalize(pathname.replace(/^\//, ''))
+  }
```
chicoxyzzy: I think that this is not correct as `os.platform()` is either a non-nullable or throws an exception that undefined (`platform`) is not a function. Probably we can just use `path.normalize` in the `exists` function below

chicoxyzzy (approved):

PR #331, opened 2023-04-11
Support android for 'ssc list-devices'

### LINKED ISSUE #319: ssc list-devices should support Android on all build hosts
https://github.com/socketsupply/socket/blob/master/src/cli/cli.cc#L1398

jwerle:
Fixes #319

### DIFF
--- a/npm/src/index.js
+++ b/npm/src/index.js
@@ -36,7 +36,7 @@ export const firstTimeExperienceSetup = async () => {
 
   let platform = ''
   process.argv.slice(2).forEach(arg => {
-    if (arg === 'setup') {
+    if (arg === 'setup' || arg === 'build') {
       isSetupCall = true
     }
 
@@ -63,7 +63,7 @@ export const firstTimeExperienceSetup = async () => {
       ['.\\bin\\install.ps1', `-fte:${platform.length > 0 ? platform : 'all'}`],
       startInfo
     )
-  } else if (platform && platform !== 'android') {
+  } else if (isSetupCall && platform && platform !== 'android') {
     // only android setup supported here, don't attempt to run fte
     spawnArgs.push(
       // @ts-ignore
--- a/npm/src/index.js
+++ b/npm/src/index.js
@@ -3,7 +3,9 @@ import os from 'node:os'
 import fs from 'node:fs'
 import { spawn } from 'node:child_process'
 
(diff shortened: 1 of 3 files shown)


src/cli/templates.hh
```diff
@@ -59,13 +59,13 @@ ssc v{{ssc_version}}
...
 
 options:
-  --platform  any : ios, android
-  --ecid      show device ECID
-  --udid      show device UDID
-  --only      only show ECID or UDID of the first device
+  --platform    android|ios
+  --ecid        show device ECID
+  --udid        show device UDID
+  --only        only show ECID or UDID of the first device
```
chicoxyzzy, line 68: lets add a note that these 3 are iOS-only commands

npm/src/index.js
```diff
@@ -36,7 +36,7 @@ export const firstTimeExperienceSetup = async () => {
 
   let platform = ''
   process.argv.slice(2).forEach(arg => {
-    if (arg === 'setup') {
+    if (arg === 'setup' || arg === 'build') {
```
mribbons, line 39: What is the reason for this change? I think it might trigger an FTE run for every build.
This isn't good if the user has elected to install android and not windows - We don't want to prompt them to install windows deps again.

src/cli/cli.cc
```diff
@@ -1394,6 +1398,21 @@ int main (const int argc, const char* argv[]) {
         log("Could not list devices using " + command);
         exit(1);
       }
+    } else if (targetPlatform == "android") {
+      auto r = exec("adb devices | tail -n +2");
```
mribbons, line 1402: This expects `adb` to be in `$PATH`. It looks as though you have implemented it correctly below.

mribbons (changes requested):

chicoxyzzy (approved):

src/cli/cli.cc
mribbons: This change could be covered in Android CI by testing for at least 1 connected emulator.

mribbons (changes requested):

src/cli/cli.cc
jwerle: oops good catch!

jwerle: Can come back to that type of test

mribbons (approved):

Issue #348, opened 2023-04-13
MacOS - Android FTE Appears to hang at # Ensuring Android dependencies are installed


mribbons:
```
  inflating: gradle-8.0.2/lib/plugins/plexus-cipher-1.7.jar  
  inflating: gradle-8.0.2/lib/plugins/plexus-interpolation-1.26.jar  
  inflating: gradle-8.0.2/lib/plugins/plexus-utils-3.3.0.jar  
  inflating: gradle-8.0.2/lib/plugins/dd-plist-1.21.jar  
  inflating: gradle-8.0.2/lib/plugins/snakeyaml-1.32.jar  
  inflating: gradle-8.0.2/lib/plugins/gradle-wrapper-8.0.2.jar  
  inflating: gradle-8.0.2/lib/plugins/gradle-java-compiler-plugin-8.0.2.jar  
  inflating: gradle-8.0.2/lib/plugins/ion-java-1.0.2.jar  
  inflating: gradle-8.0.2/lib/plugins/opentest4j-1.2.0.jar  
  inflating: gradle-8.0.2/lib/plugins/opencensus-api-0.31.1.jar  
  inflating: gradle-8.0.2/lib/plugins/grpc-context-1.27.2.jar  
Do you want to automatically accept all Android SDK Manager licenses? [y/N]
> y
# Ensuring Android dependencies are installed
```

luismiddleton:
Can confirm the same error when attempting to run `pnpm create socket-app`.
### Specs
```
System Version: macOS 13.3.1 (22E261)
Kernel Version: Darwin 22.4.0
```

mribbons:
Hi @luismiddleton,
Try this:
```bash
export SOCKET_HOME=$(ssc env | grep SOCKET_HOME)
vi $SOCKET_HOME/.ssc.env
# Remove the JAVA_HOME line if the value is '/usr', save the file
ssc setup --platform=android
```

luismiddleton:
> Hi @luismiddleton, Try this:
> 
> ```shell
> eval "export $(ssc env | grep SOCKET_HOME)"
> vi $SOCKET_HOME/.ssc.env
> # Remove the JAVA_HOME line if the value is '/usr', save the file
> ssc setup --platform=android
> ```

@mribbons  Not having any luck, it appears to still hang (even after removing the line for `JAVA_HOME` and saving the `.env`):
```bash
[...]

PR #354, opened 2023-04-16
darwin / android fte: fix usr javac hang

mribbons:
android fte/darwin: Don't use /usr/bin/javac
    
fixes https://github.com/socketsupply/socket/issues/348

### DIFF
--- a/bin/android-functions.sh
+++ b/bin/android-functions.sh
@@ -95,6 +95,12 @@ function test_javac_version() {
   local target_version=$2
   local jc_v
 
+  # Setting JAVA_HOME to /usr causes sdkmanager to hang
+  if [[ "$javac" == "/usr/bin/javac" ]] && [[ "$(host_os)" == "Darwin" ]]; then
+    write_log "h" "warn - Ignoring system java selector ($javac)"
+    return 1
+  fi
+
   jc_v="$("$javac" --version 2>/dev/null)"; r=$?
   if [[ "$r" != "0" ]]; then
     return $r
--- a/bin/android-functions.sh
+++ b/bin/android-functions.sh
@@ -204,9 +204,9 @@ function get_android_paths() {
 
     if ! test -d "$_jh" ; then
       write_log "d" "# find $java_home_test -type f -name "javac$exe"' -print0 2>/dev/null | while IFS= read -r -d '' javac"
-      find "$java_home_test" -type f -name "javac$exe" -print0 2>/dev/null | while IFS= read -r -d '' javac; do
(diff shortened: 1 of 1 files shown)


bin/android-functions.sh
```diff
@@ -95,6 +95,12 @@ function test_javac_version() {
   local target_version=$2
   local jc_v
 
+  # Setting JAVA_HOME to /usr causes sdkmanager to hang
+  if [[ "$javac" == "/usr/bin/javac" ]] && [[ "$(host_os)" == "Darwin" ]]; then
```
jwerle, line 99: is `/usr/bin/javac` the real path? what does `realpath -f /usr/bin/javac` say?

chicoxyzzy (approved):

bin/android-functions.sh
mribbons: Are you thinking of `readlink`?

```bash
% readlink -f /usr/bin/javac
/usr/bin/javac
```

mribbons: I think it's there to help people run `java`/`javac` when they aren't in `$PATH`. If you have `$JAVA_HOME` set, it runs that version for you.

jwerle: No I meant realpath, but readlink will do it too.

I am curious why this doesn't work:

```sh
java -XshowSettings:properties -version 2>&1 > /dev/null | grep 'java.home' 
```

mribbons: I haven't tried that, how should it be used?

mribbons: Thinking about this more, the reason it hangs is probably because `/usr/bin/java` goes into an infinite loop of trying to run itself.
`/usr/bin/java` runs the java in `$JAVA_HOME/bin`.
If `JAVA_HOME` is `usr`, then `/usr/bin/java` is going to run itself.
We just need to avoid using `/usr/bin/java`,.
If `$JAVA_HOME` is set, we don't need it.
We don't need `java` in `$PATH`, we just need to test that `$JAVA_HOME` is valid.

jwerle: > I haven't tried that, how should it be used?

I think just like:

```sh
JAVA_HOME="$(java -XshowSettings:properties -version 2>&1 > /dev/null | grep 'java.home')"
```

mribbons: That's only going to work if there's already a `JAVA_HOME` set, and it has to be a valid version for it to be of any use.

chicoxyzzy:
are there any blockers to merge this @jwerle ?

PR #357, opened 2023-04-17
dgram: don't use localhost as addr on win32

mribbons:
Resolves a relay connectivity issue

### DIFF
--- a/api/dgram.js
+++ b/api/dgram.js
@@ -22,7 +22,6 @@ import console from './console.js'
 import ipc from './ipc.js'
 import dns from './dns.js'
 import gc from './gc.js'
-import os from './os.js'
 
 import * as exports from './dgram.js'
 
@@ -114,9 +113,7 @@ function fromBufferList (list) {
 
 function getDefaultAddress (socket) {
   if (socket.type === 'udp6') return '::1'
-  if (socket.type === 'udp4') {
-    return os.platform() === 'win32' ? '127.0.0.1' : '0.0.0.0'
-  }
+  if (socket.type === 'udp4') return '0.0.0.0'
 
   return null
 }
--- a/test/src/dgram.js
+++ b/test/src/dgram.js
@@ -6,7 +6,6 @@ import crypto from 'socket:crypto'
 import Buffer from 'socket:buffer'
 import dgram from 'socket:dgram'
 import util from 'socket:util'
-import os from 'socket:os'
 
 // node compat
 /*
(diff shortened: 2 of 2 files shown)


api/dgram.js
```diff
@@ -114,9 +113,7 @@ function fromBufferList (list) {
 
 function getDefaultAddress (socket) {
   if (socket.type === 'udp6') return '::1'
-  if (socket.type === 'udp4') {
-    return os.platform() === 'win32' ? '127.0.0.1' : '0.0.0.0'
-  }
+  if (socket.type === 'udp4') return '0.0.0.0'
```
chicoxyzzy, line 116: it seems that it should always be `127.0.0.1`

jwerle: Yeah a lot of us had issues on our machines with using all interfaces

chicoxyzzy: oh I forgot to add a link
https://nodejs.org/dist/latest-v19.x/docs/api/dgram.html#socketconnectport-address-callback

mribbons: `127.0.0.1` is internal, how can it be used to connect to other peers?

If we're following the nodejs approach of defaulting to `127.0.0.1`, why are we only doing that on windows?

My issue is that the stream-relay code doesn't work on windows, because `127.0.0.1` is an internal address.

cc @heapwolf

chicoxyzzy: I don't remember the story behind that if statement. I think that's fine to use `0.0.0.0`, but we need to add a compatibility difference with Node to the docs for `socket.connect` and `socket.bind`

mribbons:
I see CI is failing.
```js
const address = os.platform() === 'win32' ? '127.0.0.1' : '0.0.0.0'
```

chicoxyzzy:
this string (6 occurrences) should be replaced with
```js
const address = '0.0.0.0'
```
in the `test/src/dgram.js`

api/dgram.js
jwerle: > `127.0.0.1` is internal, how can it be used to connect to other peers?
> 
> If we're following the nodejs approach of defaulting to `127.0.0.1`, why are we only doing that on windows?
> 
> My issue is that the stream-relay code doesn't work on windows, because `127.0.0.1` is an internal address.
> 
> cc @heapwolf

The default address should be a local, internal address. Ideally, it is `0.0.0.0` (all interfaces), but we had issues on Windows using it though so we used `127.0.0.1`. We should understand and fix this issue

mribbons: > The default address should be a local, internal address. Ideally, it is `0.0.0.0` (all interfaces), but we had issues on Windows using it though so we used `127.0.0.1`. We should understand and fix this issue

Ok, I think I have figured it out.

The client / server tests were failing on windows, apparently connecting to `0.0.0.0` doesn't work.

I modified all of these tests to bind specifically to `127.0.0.1`.

This means the default will still be to listen on `0.0.0.0`, but when making a client connection we have to specify the endpoint.

Thoughts?

chicoxyzzy (approved):

api/dgram.js
mribbons: @jwerle tests are passing, please approve?

jwerle (approved):

PR #359, opened 2023-04-18
install.sh: use .hh mtimes to invalidate .o

mribbons:
Also use VERSION.txt mtime

Resolves publish-npm-modules.sh error: "Repo $REPO_VERSION and $SOCKET_HOME/bin/ssc $BUILD_VERSION don't match."

This is a quick way of handling this without determining the actual .hh dependency chain within each .cc file


jwerle (approved):
Nice. Very needed! We should spend time soon removing implementation from header files with the exception of a few inlines

PR #366, opened 2023-04-21
Macos (Android) CI: Incorporate Android FTE

mribbons:
### Note that https://github.com/socketsupply/socket/pull/354 should be merged first, and ASAP.

+ switch to using test-android-emulator
    This allows us to test outside CI more reliably, rather than relying
    on github actions

+  Implemented timeout / watchdog setup to handle app failures
    that aren't caught by main adb loop in poll-adb-logcat.sh

+ macos android ci: Don't use avd cache

    "The cache became corrupted"
    The cache doesn't seem any faster than just firing up a clean AVD image

+ fix jniLibs not building when flagRunUserBuildOnly set (CI wasn't testing this before, libsocket-runtime.so wasn't being deployed with the app)

+ implement app start timeout and error output to catch failed app starts, which in turn prevents CI hanging because it would never receive a signal from webview js

+ dont search /Applications for java, this is extremely slow under CI

+ require exact java version
   
android BUG! exception in phase 'semantic analysis' in source unit '_BuildScript_' Unsupported class file major version 64

+ only build x86_64 - performance

+ determine_cxx: fix CXX not persisted to .ssc.env

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,35 +198,7 @@ jobs:
         SSC_ANDROID_CI: true
         ANDROID_SUPPORTED_ABIS: x86_64
         NODE_AUTH_TOKEN: ${{ secrets.PACKAGE_PAT }}
-
-    - name: AVD cache
-      uses: actions/cache@v3
-      id: avd-cache
-      with:
-        path: |
-          ~/.android/avd/*
-          ~/.android/adb*
-        key: avd-33
-
-    - name: create AVD and generate snapshot for caching
-      if: steps.avd-cache.outputs.cache-hit != 'true'
-      uses: reactivecircus/android-emulator-runner@v2
-      with:
-        api-level: 33
-        arch: x86_64
-        disable-animations: false
-        emulator-boot-timeout: 1200
-        emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-        force-avd-creation: false
-        ndk: 25.0.8775105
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -174,78 +174,45 @@ jobs:
...
+    - name: Build Socket Runtime
+      run: |
+        ./bin/install.sh
+        ./bin/ci_version_check.sh
       env:
+        VERBOSE: 1
+        NO_IOS: true
         CI: true
         SSC_ANDROID_CI: true
+        ANDROID_HOME:
```
chicoxyzzy, line 200: is it intended that it's empty?

.github/workflows/ci.yml
```diff
@@ -174,78 +174,45 @@ jobs:
...
-          ANDROID_SUPPORTED_ABIS="x86_64" npm run test:android
+          npm run test:android-emulator
       env:
-        CI: true
+        VERBOSE: 1
+        NO_IOS: 1
+        CI: 1
         SSC_ANDROID_CI: true
+        ANDROID_SUPPORTED_ABIS: x86_64
+        JAVA_HOME: ""
```
chicoxyzzy, line 214: hmm... maybe we need some comment here

mribbons: Have added comments. 
These are definitely required. We need to override the CI provided values.
We want to use the `.ssc.env` values, but env vars take precedence, in this case that is bad.

jwerle (approved):

PR #367, opened 2023-04-21
Typescript typings

jwerle:
This improves DX quite a bit!

image

See [`api/index.d.ts](https://github.com/socketsupply/socket/blob/459c360d937dcfbddf4721bdb4a804605b7a02f0/api/index.d.ts)


tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "lib": ["es2022", "dom"],
+
+    "emitDeclarationOnly": true,
+    "strictNullChecks": false,
+    "declarationMap": false,
+    "alwaysStrict": true,
+    "declaration": true,
+    "sourceMap": false,
+    "checkJs": false,
+    "allowJs": true
```
chicoxyzzy, line 17: It's also possible to use `jsconfig.json`, but either should work
https://code.visualstudio.com/docs/languages/jsconfig

chicoxyzzy (approved):

mribbons (approved):

package.json
```diff
@@ -4,8 +4,10 @@
     "api"
   ],
   "scripts": {
-    "pretest": "standard .",
+    "lint": "standard . && tsc --noEmit",
     "readme": "node ./bin/generate-docs.js",
+    "tsc": "./bin/generate-typescript-typings.sh",
+    "pretest": "npm run lint",
```
chicoxyzzy, line 10: I did somewhat similar in https://github.com/socketsupply/socket/pull/368, but also removed `pretest`. I think we don't need this on pretest as well as `standard .` and `gen-doc` (aka `readme`). Anyway, I can rebase after you merge it.

PR #368, opened 2023-04-21
chore(ci): fail early

chicoxyzzy:
Fail CI early when linting doesn't pass or when git working directory is not clean after generating docs. This can save resources on building ssc on the platforms.


heapwolf (approved):

mribbons (approved):

mribbons:
Awesome

PR #370, opened 2023-04-22
Minor FTE improvements

chicoxyzzy:
- add a `git` dependency which is not pre-installed on Ubuntu
- add `dnf` which is used on Fedora by default now

### DIFF
--- a/bin/functions.sh
+++ b/bin/functions.sh
@@ -561,6 +561,7 @@ function first_time_experience_setup() {
       if [[ "$package_manager" == "apt install" ]]; then
         log_and_run sudo apt update || return $?
         log_and_run sudo apt install -y     \
+          git                   \
           libwebkit2gtk-4.1-dev \
           build-essential       \
           libc++abi-14-dev      \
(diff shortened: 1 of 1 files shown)


bin/functions.sh
```diff
@@ -571,15 +572,32 @@ function first_time_experience_setup() {
...
         log_and_run sudo pacman -Syu \
+          git            \
           webkit2gtk-4.1 \
           base-devel     \
           libc++abi-14   \
           libc++1-14     \
           clang-14       \
           pkgconf        \
           || return $?
+          exit $?
```
chicoxyzzy, line 583: this one was missing

bin/functions.sh
```diff
@@ -561,25 +561,29 @@ function first_time_experience_setup() {
...
         log_and_run sudo apt install -y     \
+          git                   \
           libwebkit2gtk-4.1-dev \
           build-essential       \
           libc++abi-14-dev      \
           libc++-14-dev         \
           pkg-config            \
           clang-14              \
           || return $?
-          exit $?
```
chicoxyzzy, line 571: I think it's not necessary here since we have `return` above?

heapwolf (approved):

jwerle (approved):

mribbons (approved):

PR #380, opened 2023-04-25
0 2 12

heapwolf:


getify (changes requested):
I may be missing this, but I don't see some of the scheme:PTP changes brought in here that were landed on the `stream-relay-protocol`: `main` branch.

I think there were changes to `scheme-ptp.js`, and possibly to `packets.js`.

heapwolf:
> I may be missing this, but I don't see some of the scheme:PTP changes brought in here that were landed on the `stream-relay-protocol`: `main` branch.
> 
> I think there were changes to `scheme-ptp.js`, and possibly to `packets.js`.

you're right, I rebased this branch so should contain this stuff now

Issue #374, opened 2023-04-24
Getting errors when I run `ssc build -r --platform=ios-simulator` 


lamiazar:
I'm getting errors when I try to run `ssc build -r --platform=ios-simulator`:
```
ssc build -r --platform=ios-simulator
• WARNING: $CXX env var not set, assuming defaults +12ms
• cleaned: /Users/lamiabedjou/socket/build/ios-simulator +15ms
• package prepared +148ms
• using '/Users/lamiabedjou/.local/share/socket/' as 'SOCKET_HOME' +4ms
• building for iOS +9ms
• ERROR: failed to archive project +1310ms
• Command line invocation:
    /Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild build CONFIGURATION_BUILD_DIR=/Users/lamiabedjou/socket/build/ios-simulator -configuration Debug -scheme beepboop-dev -destination "platform=iOS Simulator,OS=latest,name=iPhone 14" CODE_SIGN_IDENTITY= CODE_SIGNING_REQUIRED=NO CODE_SIGN_ENTITLEMENTS= CODE_SIGNING_ALLOWED=NO

User defaults from command line:
    IDEPackageSupportUseBuiltinSCM = YES

Build settings from command line:
    CODE_SIGN_ENTITLEMENTS = 
    CODE_SIGN_IDENTITY = 
    CODE_SIGNING_ALLOWED = NO
    CODE_SIGNING_REQUIRED = NO
    CONFIGURATION_BUILD_DIR = /Users/lamiabedjou/socket/build/ios-simulator

Prepare packages

note: Using codesigning identity override: 
Computing target dependency graph and provisioning inputs

Create build description
Build description signature: 7277b6885f5920636f8c333e59281c4f
Build description path: /Users/lamiabedjou/Library/Developer/Xcode/DerivedData/beepboop-dev-chuxgglboincwvcbytyrmtjhrwzr/Build/Intermediates.noindex/XCBuildData/7277b6885f5920636f8c333e59281c4f-desc.xcbuild

note: Building targets in dependency order
MkDir /Users/lamiabedjou/socket/build/ios-simulator/beepboop-dev.app (in target 'beepboop-dev' from project 'beepboop-dev')
    cd /Users/lamiabedjou/socket/build/ios-simulator
    /bin/mkdir -p /Users/lamiabedjou/socket/build/ios-simulator/beepboop-dev.app

[...]

PR #384, opened 2023-04-26
arm64-iPhoneSimulator: Enable support

mribbons:
Fixes silent crash when testing apps on iPhoneSimulator running on Mac M1 or later

Fixes https://github.com/socketsupply/socket/issues/374

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -2331,7 +2331,7 @@ int main (const int argc, const char* argv[]) {
       }
 
       if (!fs::exists(deviceObjects)) {
-        log("ERROR: libs folder for the target platform doesn't exist: " + deviceObjects.string());
+        log("ERROR: objects folder for the target platform doesn't exist: " + deviceObjects.string());
       }
 
       if (!fs::exists(deviceLibs) || !fs::exists(deviceObjects)) {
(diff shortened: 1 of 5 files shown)


chicoxyzzy (approved):
works on x86_64

src/cli/cli.cc
```diff
@@ -2321,32 +2321,35 @@ int main (const int argc, const char* argv[]) {
...
+      if (!fs::exists(deviceLibs)) {
+        log("ERROR: libs folder for the target platform doesn't exist: " + deviceLibs.string());
+      }
+
+      if (!fs::exists(deviceObjects)) {
+        log("ERROR: libs folder for the target platform doesn't exist: " + deviceObjects.string());
       }
 
+      if (!fs::exists(deviceLibs) || !fs::exists(deviceObjects)) {
+        exit(1);
```
chicoxyzzy, line 2338: can we add error message here?

mribbons: There are error messages a few lines above. I have combined the exists check into one if statement so that both errors can be reported before terminating.
```cc
      if (!fs::exists(deviceLibs)) {
        log("ERROR: libs folder for the target platform doesn't exist: " + deviceLibs.string());
      }

      if (!fs::exists(deviceObjects)) {
        log("ERROR: objects folder for the target platform doesn't exist: " + deviceObjects.string());
      }

      if (!fs::exists(deviceLibs) || !fs::exists(deviceObjects)) {
        exit(1);
      }
```

PR #392, opened 2023-04-27
npm run gen:docs

mribbons:


npm/packages/@socketsupply/socket-node/API.md
```diff
@@ -1,5 +1,5 @@
 
-### [`send(options)`](https://github.com/socketsupply/socket/blob/master/npm/packages/@socketsupply/socket-node/index.js#L168)
+### [`send(options)`](https://github.com/socketsupply/socket/blob/master/npm\packages\@socketsupply\socket-node\index.js#L168)
```
jwerle, line 2: is this right?

mribbons: Ran it on windows, so that might explain why so many changes appeared.
Will run again on macos and see if it all goes away.

mribbons: Ok, on macos gen:docs shows no changes against master, will can this one :pray:

PR #394, opened 2023-04-28
Fix `ssc run` to be on par with `ssc build`

chicoxyzzy:
Next step is to make things DRY for `build` and `run`. Probably we can just call `run` from `build` filtering and passing arguments

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -250,7 +250,7 @@ void signalHandler (int signal) {
   appMutex.unlock();
 }
 
-int runApp (const Path& path, const String& args, bool headless) {
+int runAppCommon (const Path& path, const String& args, bool headless) {
   auto cmd = path.string();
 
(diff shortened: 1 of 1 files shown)


mribbons (approved):

chicoxyzzy:
I just found a few more issue with `run` :(
So I'm going to separate `build -r` / `run` to a separate step in this PR to avoid further divergence

chicoxyzzy:
@mribbons @jwerle I can't make it work after rebase. I've tried a few times from scratch, but my knowledge is not enough. Can you halp me to move all the run stuff from in the `ssc build` inside of the latest `if (flagShouldRun) {` body so we have only one such condition there?

chicoxyzzy:
Otherwise I can't share that code between `ssc build` and `ssc run` later

chicoxyzzy:
replaced by #409 

Further changes will be in upcoming PRs

PR #397, opened 2023-04-28
Clean up useless API imports

### LINKED ISSUE #386: Remove `console` dependency from `crypto` module
As far as I can tell, the [`console` module is a dependency](https://github.com/socketsupply/socket/blob/b11f8ab9e9cba3093597836c08c32a90e6b150e1/api/crypto.js#L15) of the `crypto` module only [for one single `console.warn(..)` line](https://github.com/socketsupply/socket/blob/b11f8ab9e9cba3093597836c08c32a90e6b150e1/api/crypto.js#L83):

```js
 console.warn('Missing implementation for globalThis.crypto.getRandomValues()')
```

This seems like an unnecessary warning message to report, as I am pretty sure we're only supporting webviews that will have this -- if the webview has `globalThis`, which is relatively new to JS in the last couple of years, it's fairly modern, and thus will almost certainly have long before that added the crypto stuff.

----

To the specific point of this issue, the `console` module relies on the `ipc` module to marshal its messages across the IPC boundary. But the crypto module really should be able to be used even without IPC -- like directly in a browser. If we get rid of this one `console.warn()` message... or even just leave it but don't attempt to marshal it with IPC... then we free up the `crypto` module from such a (unnecessary, IMO) dependency, making it more flexible in its use.

### LINKED ISSUE #387: Remove `process` dependency from `libsodium.js` file
It seems like [this `process` dependency in `libsodium.js`](https://github.com/socketsupply/socket/blob/b11f8ab9e9cba3093597836c08c32a90e6b150e1/api/external/libsodium/index.js#L1) is unnecessary and could instead be replaced with:

```js
const process = {}
```

IOW, it doesn't seem like libsodium actually relies on anything important from node's `process`. If that's indeed true, we should remove this dependency.

jwerle:
Fixes #386
Fixes #387

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -320,7 +320,7 @@ Start advertising a new value for a well-known UUID
 | Not specified | Promise<void> |  |
 
 
-# [Crypto](https://github.com/socketsupply/socket/blob/master/api/crypto.js#L13)
+# [Crypto](https://github.com/socketsupply/socket/blob/master/api/crypto.js#L14)
 
 
  Some high-level methods around the `crypto.subtle` API for getting
--- a/api/crypto.js
+++ b/api/crypto.js
@@ -1,3 +1,4 @@
+/* global console */
 /**
  * @module Crypto
  *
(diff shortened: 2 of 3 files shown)


getify (approved):
LGTM, thanks!

Issue #363, opened 2023-04-18
The terminal shows bugs after the app builds on ubuntu with `curl`


lamiazar:
@mribbons I was able to get `ssc` installed with the new `curl` command 
the app window opens shows with `ssc build -r -o`, but I see some bugs printed on the terminal
```
ok - done. type 'ssc -h' for help
mia@mia:~/socket/socket$ ssc -v
0.2.1 (bd6232e7)
mia@mia:~/socket/socket$ mkdir app
mia@mia:~/socket/socket$ cd app
mia@mia:~/socket/socket/app$ ssc init
• socket.ini created in /home/mia/socket/socket/app +6ms
• src/index.html created in /home/mia/socket/socket/app +0ms
• .gitignore created in /home/mia/socket/socket/app +0ms
mia@mia:~/socket/socket/app$ ssc build -r -o
• preparing build for linux +3ms
• WARNING: [linux] icon '/home/mia/socket/socket/app/src/icon.png' does not exist +11ms
• package prepared +0ms
• compiled native binary +3768ms
• Running App: /home/mia/socket/socket/app/build/linux/beepboop-dev_v1.0.0-1_amd64/opt/beepboop-dev/beepboop-dev --from-ssc +0ms

(beepboop-dev:21591): Gdk-CRITICAL **: 10:02:15.868: gdk_monitor_get_workarea: assertion 'GDK_IS_MONITOR (monitor)' failed

• Creating Window#0 +0ms
• Showing Window#0 (seq=) +677ms
*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug


*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug


*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug


*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug


*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug



(beepboop-dev:21591): Gtk-WARNING **: 10:03:12.412: Negative content width -6 (allocation 18, extents 12x12) while allocating gadget (node label, owner GtkLabel)

^CSegmentation fault (core dumped)
```

lamiazar:
[...]

PR #398, opened 2023-04-28
Set monitor default geometry on Linux

jwerle:
https://github.com/socketsupply/socket/issues/363

### DIFF
--- a/src/window/linux.cc
+++ b/src/window/linux.cc
@@ -4,6 +4,9 @@ static GtkTargetEntry droppableTypes[] = {
   { (char*) "text/uri-list", 0, 0 }
 };
 
+#define DEFAULT_MONITOR_WIDTH 720
+#define DEFAULT_MONITOR_HEIGHT 364
+
 namespace SSC {
   struct WebViewJavaScriptAsyncContext {
     IPC::Router::ReplyCallback reply;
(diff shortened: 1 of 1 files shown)


src/window/linux.cc
```diff
@@ -633,9 +641,25 @@ namespace SSC {
...
 
-      gdk_monitor_get_workarea(monitor, &geometry);
-      width = (int) geometry.width;
-      height = (int) geometry.height;
+      if (monitor) {
+        gdk_monitor_get_workarea(monitor, &geometry);
+      }
+
+      if (geometry.width > 0) {
+        width = (int) geometry.width;
```
chicoxyzzy, line 649: it is `int` already https://docs.gtk.org/gdk3/struct.Rectangle.html

chicoxyzzy (approved):

src/window/linux.cc
```diff
@@ -4,6 +4,9 @@ static GtkTargetEntry droppableTypes[] = {
   { (char*) "text/uri-list", 0, 0 }
 };
 
+#define DEFAULT_MONITOR_WIDTH 720
+#define DEFAULT_MONITOR_HEIGHT 364
```
chicoxyzzy, line 8: actually we should take these values from `windowManager->options.default{Width|height}`

lamiazar:
Screenshot from 2023-04-28 13-06-40

src/window/linux.cc
jwerle: That won't work if `[window] height = 50%` is set

lamiazar (approved):
This is now working-I was running the wrong command from the local checkout

src/window/linux.cc
chicoxyzzy: What was the issue with percents? I don't remember. all values should be in pixels in the end

PR #403, opened 2023-05-01
Fix CommonJS on Android

jwerle:


api/module.js
```diff
@@ -158,9 +178,7 @@ export const COMMONJS_WRAPPER = CommonJSModuleScope
  * @type {string}
  */
 export const MAIN_SOURCE_URL = (
-  globalThis.origin === 'file://' && globalThis.location?.href
-    ? globalThis.location.href
-    : `file://${process.cwd() || ''}`
+  globalThis.location?.href || `file://${process.cwd() || ''}`
```
mribbons, line 181: `process.cwd()` can change.. I'm not sure if an import could be triggered after user space code has modified though cwd though.... Maybe a userspace import could modify it, which would be a bad idea, but:
What about storing the starting cwd on app start, and store this on process? Accessible via `process.startDir()` perhaps.

test/src/path.js
```diff
@@ -89,16 +89,18 @@ test('path.win32.join', (t) => {
 test('path.win32.dirname', (t) => {
   t.equal(path.win32.dirname('a\\b\\c'), 'a\\b', 'a\\b')
   t.equal(path.win32.dirname('a\\b\\c\\d.js'), 'a\\b\\c', 'a\\b\\c')
+  t.equal(path.win32.dirname('C:\\a\\b\\c\\d.js'), 'C:\\a\\b\\c', 'C:\\a\\b\\c')
```
mribbons, line 92: 👍

mribbons (commented):
I trust you saw that `CommonJS` is failing on windows now...

api/module.js
chicoxyzzy: we don't implement `process.chdir()` method and `process.cwd()` is always returns `primordials.cwd` which we can use here as well

api/peer.js
```diff
@@ -137,9 +137,9 @@ export class Peer extends EventEmitter {
       this._emit(packet.usr1, packet.message, address, port)
     }
 
-    this.#peerSubscribe("onData", "data")
-    this.#peerSubscribe("onClusterData", "cluster-data")
-    this.#peerSubscribe("onClose", "close")
```
chicoxyzzy, line 142: was `peer.js` it not in the list of files to lint?

chicoxyzzy (approved):

PR #405, opened 2023-05-02
crypto: simplify nodejs webcrypto

missinglink:
does this make sense to clean up nodejs support for webcrypto?

Screenshot 2023-05-02 at 14 15 03
Screenshot 2023-05-02 at 14 15 15

### DIFF
--- a/api/crypto.js
+++ b/api/crypto.js
@@ -16,7 +16,11 @@ import console from './console.js'
 
 import * as exports from './crypto.js'
 
-let getRandomValuesFallback = null
+/**
+ * WebCrypto API
+ * @see {https://developer.mozilla.org/en-US/docs/Web/API/Crypto}
+ */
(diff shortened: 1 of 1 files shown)


missinglink:
The motivation for this was that I wanted to use a different function `webcrypto.digest` and it seemed silly to create a fallback function for each

missinglink:
note: I changed a constant `export const webcrypto` to a variable `export let webcrypto`

jwerle:
> In fact, is it even required to call `import('node:crypto')` as everything seems to be available in `globalThis`?
> 
> ```shell
> node --version
> v18.16.0
> 
> node
> Welcome to Node.js v18.16.0.
> Type ".help" for more information.
> > globalThis.crypto.webcrypto.subtle
> SubtleCrypto {}
> ```

That is only true in a REPL context

jwerle (approved):
LGTM!

PR #407, opened 2023-05-07
Android improvements

mribbons:
* android build cwd: don't run from build folder

(where possible)

Partially resolves file lock issue on windows where folders are locked
while processes are using them as pwd

Note that gradle has to run from build folder, but this change still
results in less process to kill.

* android: improve static lib dx

Always copy libs, even if jniLibs exists, this may have been an issue and the performance detriment is negligable
ERROR out if static libs don't exist in SOCKET_HOME

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -3012,51 +3012,55 @@ int main (const int argc, const char* argv[]) {
         auto libs = _main / "libs";
         auto obj = _main / "obj";
 
-        if (!fs::exists(jniLibs)) {
+        if (fs::exists(obj)) {
+          fs::remove_all(obj);
+        }
 
-          if (fs::exists(obj)) {
-            fs::remove_all(obj);
-          }
-
-          // TODO(mribbons) - Copy specific abis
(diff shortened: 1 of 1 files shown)


src/cli/cli.cc
```diff
@@ -3029,38 +3029,58 @@ int main (const int argc, const char* argv[]) {
...
 
+        auto buildJniLibs = true;
+        if (fs::exists(jniLibs)) {
+          int androidSharedLibCount = 0;
+          for (auto const& dir_entry : fs::recursive_directory_iterator(jniLibs)) {
+            if (dir_entry.path().stem() == "libsocket-runtime.so") {
+              androidSharedLibCount++;
+            }
+          }
+
```
jwerle, line 3053: Can you explain what this is checking more explicitly in an inline comment?

jwerle (changes requested):
LGTM! Just left a comment for a request for comments

chicoxyzzy (approved):
LGTM

let me know if you need any help with resolving conflicts

PR #413, opened 2023-05-08
fix(src/cli/templates.hh): fix default viewport in HTML

chicoxyzzy:


jwerle (approved):
nice! needed this for a while 😅

PR #417, opened 2023-05-09
feat(bin/functions.sh): add Fedora support

chicoxyzzy:


bin/functions.sh
```diff
@@ -579,9 +579,13 @@ function first_time_experience_setup() {
           clang-14       \
           pkgconf        \
           || return $?
-      elif [[ "$packege_manager" == "dnf install" ]]; then
-        echo "warn - dnf package manager is not suppored yet. Please try to install from npm or from source."
-        exit 1
+      elif [[ "$package_manager" == "dnf install" ]]; then
+        log_and_run sudo dnf install \
```
jwerle, line 583: Can we clean this to look like:

```sh
log_and_run sudo dnf install   \
  make                         \
  automake                     \
  gcc                          \
  gcc-c++                      \
  kernel-devel                 \
  clang14-devel                \
  clang14-libs                 \
  libcxx-devel                 \
  libcxxabi-devel              \
  webkit2gtk4.1-devel          \
  || return $?
```

jwerle (changes requested):

jwerle (approved):

mribbons (approved):

PR #418, opened 2023-05-10
OTA (prep) / Reload support

mribbons:
OTA (prep) / Reload support

This PR has been cut back due to https://github.com/socketsupply/socket/pull/443 fixing most of the issues I was attempting to resolve, however we still need the following:

Android
1. `Window.load()` - Copy all .html assets to /files so they can be
modified by initSocketPreload
2. `WebViewClient.shouldInterceptRequest()` - attempt to fulfill asset
requests from /files if they exist, then fall back to /assets, this
enables OTA updates, because /assets is read only.

https://github.com/socketsupply/socket/pull/438 was PRd against this branch instead of `master`.


api/preload.js
```diff
@@ -0,0 +1,45 @@
+import fs from 'socket:fs/promises'
+import Path from 'socket:path'
```
jwerle, line 2: typically, this is called `path`, not `Path`.

api/preload.js
```diff
@@ -0,0 +1,45 @@
...
+const preloadJavascript = `
+/// PRELOAD_JS_PLACEHOLDER ///
+`
+export const getPreloadJavascript = () => {
+  return preloadJavascript
+}
+
+const preloadScriptModule = `<script type="module">\n${preloadJavascript}\n</script>\n`
+
+const recursePath = async (path, fileFunc, data) => {
```
jwerle, line 14: Please use `function`, instead of arrow functions. It is easier to identity a function instead of a variable.

Can you make this function a little easier to read to?

```js
/**
 * @param {string} directory - Directory tree to walk down
 * @param {(string, any?): any} callback - Callback to invoke for each entry in directory tree
 * @return {Promise}
 */
async function walk (directory, callback) {
  const entries = await fs.readdir(directory, { withFileTypes: true })
  for (const entry of entries) {
    const filename = path.join(directory, entry.name)
    if (entry.isDirectory()) {
      await walk(filename, callback)
    } else {
      await callback(filename)
    }
  }
}
```

src/android/webview.kt
```diff
@@ -1,16 +1,27 @@
 // vim: set sw=2:
 package __BUNDLE_IDENTIFIER__
 import java.lang.ref.WeakReference
+import java.io.File
```
jwerle, line 4: please don't import these 🙏🏼

src/android/webview.kt
```diff
@@ -1,16 +1,27 @@
...
+import android.util.Log
 
 fun decodeURIComponent (string: String): String {
   val normalized = string.replace("+", "%2B")
   return java.net.URLDecoder.decode(normalized, "UTF-8").replace("%2B", "+")
 }
 
 fun isAssetUri (uri: android.net.Uri): Boolean {
+  for (item in uri.pathSegments) {
```
jwerle, line 13: should this be here?

api/preload.js
```diff
@@ -0,0 +1,45 @@
+import fs from 'socket:fs/promises'
+import Path from 'socket:path'
+import process from 'socket:process'
+
+const preloadJavascript = `
+/// PRELOAD_JS_PLACEHOLDER ///
```
jwerle, line 6: Can you replace

```
/// PRELOAD_JS_PLACEHOLDER ///
```

with

```
__PRELOAD_JS_PLACEHOLDER__
```

src/android/webview.kt
```diff
@@ -56,6 +71,10 @@ open class WebViewClient (activity: WebViewActivity) : android.webkit.WebViewCli
       return false
     }
 
+    if (url.scheme == "reload") {
```
jwerle, line 74: Lets just reuse `ipc://` instead as this introduces possibly another CSP entry.

Lets use

```
ipc://reload
``` 

or something like it

src/android/window.kt
```diff
@@ -89,6 +103,10 @@ open class Window (runtime: Runtime, activity: MainActivity) {
         }
       }
     })
+
+    this.activity.get()?.client?.assetManager = runtime.configuration.assetManager
```
jwerle, line 107: These 3 lines seem brittle. We should have a method on the `WebViewClient` that sets these. The `WebViewClient` has access to the `WebViewActivity` so it should at least be able to set the `rootDirectory`

src/window/win.cc
```diff
@@ -649,6 +649,8 @@ namespace SSC {
     SetWindowLongPtr(window, GWLP_USERDATA, (LONG_PTR) this);
 
     SSC::String preload = createPreload(opts);
+    // store preload JS on window because, preload variable above goes out of scope by the time browser handlers are called.
+    this->preloadJavascript = createPreload(opts);
```
jwerle, line 653: Seems like a waste to call `createPreload(opts)` twice - just do `this->preloadJavascript = preload`

src/window/win.cc
```diff
@@ -947,12 +949,17 @@ namespace SSC {
...
                               String headers;
                               char* body;
+                              String moduleSource;
 
-                              auto moduleUri = "file://" + replace(path.string(), "\\\\", "/");
-                              auto moduleSource = trim(tmpl(
-                                moduleTemplate,
-                                Map { {"url", String(moduleUri)} }
-                              ));
+                              if (uri_s.compare("preload") == 0) {
```
jwerle, line 954: what does this URI look like?

src/window/win.cc
```diff
@@ -947,12 +949,17 @@ namespace SSC {
...
+                                moduleSource = replace(readFile(path), "/// PRELOAD_JS_PLACEHOLDER ///", replace(w->preloadJavascript, "`", "\\`"));
```
jwerle, line 955: use `__PRELOAD_JS_PLACEHOLDER__` instead of `/// PRELOAD_JS_PLACEHOLDER ///`

jwerle:
> returns the size of the preload js module import, so that live reload
> detectors can account for the change in size

I don't think this will be enough to diff a change - consider a change of`abc` -> `abd`, which is the same length in characters

api/preload.js
```diff
@@ -0,0 +1,45 @@
...
+  const destDir = ''
+
+  await recursePath(process.cwd(), async (file) => {
+    if (!(file.toLowerCase().endsWith('.html') || file.toLowerCase().endsWith('.htm'))) {
+      return
+    }
+    const sourcePath = sourceDir.length > 0 ? Path.join(sourceDir, file) : file
+    const destPath = destDir.length > 0 ? Path.join(destDir, file) : file
+    let html = await fs.readFile(sourcePath)
+    if (html.indexOf('const preloadJavascript') < 0) {
```
jwerle, line 36: This seems too brittle

api/preload.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+}
+
+const initSocketPreload = async () => {
+  const sourceDir = ''
+  const destDir = ''
+
+  await recursePath(process.cwd(), async (file) => {
+    if (!(file.toLowerCase().endsWith('.html') || file.toLowerCase().endsWith('.htm'))) {
```
jwerle, line 30: we should only support `.html`  🙏🏼

src/window/win.cc
mribbons: The original value is:
`socket://preload/`
By the time it gets here, `socket://` and the trailing `/` have been removed.

src/android/window.kt
mribbons: Can you take a look at the changes I made, I'm not sure if it satisfies your concerns.

`rootDirectory` is on of `MainActivity`, but `WebViewClient` just thinks it's a `WebViewActivity` so it can't access it without a cast.
Also, I am having issues with the order things are initialized in, `Window` owns `getJavaScriptPreloadSource()`, but `Window` is initialized after `WebViewActivity` and consequently after `WebViewClient`.

src/android/webview.kt
mribbons: This was part of my initial attempt, we don't need it at all. Glad you spotted it.

mribbons:
> I don't think this will be enough to diff a change - consider a change of`abc` -> `abd`, which is the same length in characters

Keep in mind that we are also checking the mtime.

The size check is really just a backup for someone dropping in a backup file that doesn't get flagged because it is older than what is in the runtime folder (Windows maintains mtime on copy).

As the update check happens multiple times per second, we can't go as far as checking the content.

api/preload.js
mribbons: This was due to an issue with `path.join()`, but it turns out we don't currently need different source and dest paths anyway.

mribbons:
> we should consider a different name other than`preload.js`. It isn't obvious and conflates the purpose of `runtime-preload.hh`

`get_preload.js`, 
`socket://get_preload`?

mribbons:
> There is the **Apply Changes** feature available on Android too https://developer.android.com/studio/run#apply-changes

Do you have any info on using apk updates outside of Android Studio?

mribbons:
Hmm, ci failing, please help:
```
> gen:tsc
> ./bin/generate-typescript-typings.sh

Error: Process completed with exit code 1.
```

chicoxyzzy:
> ci failing

Rebase should help. Lint error means that documentation or TypeScript definitions need update. `npm run gen` fixes this. I improved the error here https://github.com/socketsupply/socket/pull/457

src/android/webview.kt
```diff
@@ -36,6 +37,12 @@ open class WebView (context: android.content.Context) : android.webkit.WebView(c
 open class WebViewClient (activity: WebViewActivity) : android.webkit.WebViewClient() {
   protected val activity = WeakReference(activity)
   open protected val TAG = "WebViewClient"
+  open protected var rootDirectory = ""
+
+  fun putRootDirectory(rootDirectory: String) {
+    this.rootDirectory = rootDirectory
```
jwerle, line 43: Can you make this `fun Detroit directory (...) ` instead?

src/android/webview.kt
```diff
@@ -84,6 +91,50 @@ open class WebViewClient (activity: WebViewActivity) : android.webkit.WebViewCli
...
+        val stream = java.io.PipedOutputStream()
+        kotlin.concurrent.thread {
+          stream.write(bytes, 0, bytes.size)
+          stream.close()
+        }
+
+        val response = android.webkit.WebResourceResponse(
+          if (filePath.toString().endsWith(".js")) "text/javascript" else "text/html",
+          "utf-8",
+          java.io.PipedInputStream(stream)
```
jwerle, line 123: Why not just use `input` here instead of buffer the entire stream above to use here

src/android/webview.kt
mribbons: auto correct typo, please clarify :)

api/ipc.js
```diff
@@ -950,11 +950,15 @@ export class Result {
  * @ignore
  */
 export async function ready () {
-  return await new Promise((resolve) => {
+  const startReady = new Date().getTime()
```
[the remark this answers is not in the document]
mribbons: 🙏 GTK

src/android/webview.kt
mribbons: Agree, old code from when the buffer was being modified.

src/android/webview.kt
```diff
@@ -116,22 +167,23 @@ export default module
         "Access-Control-Allow-Methods" to "*"
       )
 
+      // prevent piped streams blocking each other, have to write on a separate thread if data > 1024 bytes
       val bytes = moduleTemplate.toByteArray()
-      stream.write(bytes, 0, bytes.size)
-      stream.close()
+      kotlin.concurrent.thread {
+        stream.write(bytes, 0, bytes.size)
```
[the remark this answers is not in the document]
mribbons: The code was previously like this, the only change on this branch is to use `kotlin.concurrent.thread`. 

At one point I was sending preload js via this route, and I got stuck with thread lockups due to buffers > 1024 bytes.

I think it doesn't hurt to have `kotlin.concurrent.thread` there in case someone makes a similar modification in the future, thoughts?

src/android/webview.kt
jwerle: oops i meant 
> Can you make this fun setRootDirectory (...) instead?

chicoxyzzy (approved):

mribbons:
> LGTM - we can rename `putDirectoryRoot` to `setDirectoryRoot` in a subsequent PR so you can move forward!

This was my first choice, but kotlin seems to reserve method names starting with `get` or `set`.

PR #423, opened 2023-05-11
Improve peer packet handler

jwerle:


api/peer.js
```diff
@@ -134,7 +136,27 @@ export class Peer extends EventEmitter {
...
-      this._emit(packet.usr1, packet.message, address, port)
+      const event = packet.usr1
+      let message = packet.message
+
+      if (!event || typeof event !== 'string') return
+
+      if (isBufferLike(message)) {
+        try {
+          message = textDecoder.decode(message)
+        } catch (err) {}
```
chicoxyzzy, line 147: should it emit an error here as well?

jwerle: it may not be a valid string to decode, we still want to emit the packet message buffer

chicoxyzzy (approved):

PR #434, opened 2023-05-14
Fix generated typings

jwerle:


api/diagnostics/channels.js
```diff
@@ -146,9 +146,11 @@ export class Channel {
 
   /**
    * A no-op for `Channel` instances. This function always returns `false`.
-   * @return {boolean}
+   * @param {string} name
+   * @param {object} message
+   * @return Promiise<boolean>
```
chicoxyzzy, line 151: typo here and many other places

jwerle: looks like just 2! fixing now

api/worker.js
```diff
@@ -1,7 +1,7 @@
 import console from './console.js'
 
 // eslint-disable-next-line no-new-func
-const GlobalWorker = new Function('return Worker')()
+const GlobalWorker = new Function('return this.Worker')()
```
chicoxyzzy, line 4: why not `globalThis`?

jwerle: Because typescript 😑😅

jwerle: no :( we can't use the webworker lib without conflicts

chicoxyzzy (approved):

PR #435, opened 2023-05-15
win32 + other path related failing tests

mribbons:
Examples of nodejs behavior for the newly added tests:
```js
import process from 'node:process'
import path from 'node:path'
import fs, { constants } from 'node:fs'

// join: c:\tests\nodefs
console.log(`join: ${process.cwd()}, ${path.join(process.cwd() + '\\', `${new Date().getTime()}.log`)}`)

// cwd last char: s
console.log(`cwd last char: ${process.cwd()[process.cwd().length-1]}`)

// c:\test\hello
console.log(`join empty paths: ${path.join('c:\\test', '', '', 'hello')}`)

// c:\test\hello
console.log(`join slashes: ${path.join('c:\\test', '\\', '\\', 'hello')}`)

// doesn't throw
fs.promises.mkdir('mkdir\\test\\path', {recursive: true})
fs.promises.access('mkdir\\test\\path', constants.X_OK)

// would be great to also see copyFile and utimes working
fs.writeFileSync('text.dat', '1234')
fs.promises.copyFile('text.dat', 'text2.dat')
let yesterday = new Date(new Date().getTime() - 86400000);
fs.promises.utimes('text.dat', new Date(), yesterday)
fs.promises.stat('text.dat').then((stat) => {
  if (stat.mtimeMs === yesterday.getTime()) {
    console.log(`utimes is good`)
  } else {
    console.log(`${stat.mtime} !== ${yesterday.getTime()}`)
  }
})
```

### DIFF
--- a/test/src/fs/promises.js
+++ b/test/src/fs/promises.js
@@ -42,6 +42,13 @@ if (process.platform !== 'ios') {
       const { err } = await fs.mkdir(dirname, {})
       t.equal(err, undefined, 'mkdir does not throw')
     })
+
+    test('fs.promises.mkdir({recursive: true})', async (t) => {
+      const dirname = FIXTURES + Math.random().toString(16).slice(2)
+      const { err } = await fs.mkdir(`${dirname}${path.sep}${dirname}`, {recursive: true})
+      await fs.access(`${dirname}${path.sep}${dirname}`, X_OK)
+      t.equal(err, undefined, 'mkdir does not throw and creates directories recursively')
+    })
   }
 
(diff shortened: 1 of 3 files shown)


test/src/fs/promises.js
```diff
@@ -42,6 +42,13 @@ if (process.platform !== 'ios') {
       const { err } = await fs.mkdir(dirname, {})
       t.equal(err, undefined, 'mkdir does not throw')
     })
+
```
jwerle, line 45: This is not supported yet so can you remove this test to narrow the scope of just `path`

PR #438, opened 2023-05-15
Build static clang spaces

mribbons:
android_clang: 
return binary path and --target separately
this allows $clang calls to be quoted in all build functions


chicoxyzzy (changes requested):
please drop the v0.3.1 commit from this PR

chicoxyzzy (approved):
windows build works on this branch

jwerle (approved):

PR #441, opened 2023-05-16
Clean up internal init, fix 'webview:onload' test

jwerle:


api/internal/init.js
```diff
@@ -267,25 +267,22 @@ class RuntimeXHRPostQueue extends ConcurrentQueue {
     const options = { responseType: 'arraybuffer' }
     const result = await ipc.request('post', { id }, options)
 
+    promise.resolve()
```
chicoxyzzy, line 270: would it make sense if will resolve on `!result.err` and reject (along with dispatching an event) otherwise?

api/test/index.js
```diff
@@ -547,3 +550,18 @@ function rethrowImmediate (err) {
...
+ * @param {unknown} thing
+ * @returns {string}
+ */
+function toJSON (thing) {
+  /** @type {(_k: string, v: unknown) => unknown} */
+  const replacer = (_k, v) => (v === undefined) ? '_tz_undefined_tz_' : v
+
+  const json = JSON.stringify(thing, replacer, '  ') || 'undefined'
+  return json.replace(/"_tz_undefined_tz_"/g, 'undefined')
+}
```
chicoxyzzy, line 567: 👍

chicoxyzzy (approved):

api/internal/init.js
jwerle: we don't want to reject - this is just me moving code around to be a bit cleaner!

PR #442, opened 2023-05-16
iOS Simulator stdout

jwerle:

PR #443, opened 2023-05-16
Revert preload injection

jwerle:


chicoxyzzy:
let's merge #441 and #442 so it will be easier to review this one without changes from those two

chicoxyzzy (approved):

PR #448, opened 2023-05-17
fix(bin/android-functions.sh): fix typo

chicoxyzzy:


jwerle (commented):
I don't think this was a typo. It is the past tense of `elect`, but perhaps @mribbons can say more! 😅

mribbons:
Indeed, 'elected' isn't a typo.

We can use 'selected' if you like @chicoxyzzy, another option would be 'chosen'.

'selected' doesn't feel right to me because it reminds me of `<select>`, and this is terminal input :joy:

chicoxyzzy:
No need to alter anything; it's all good

PR #450, opened 2023-05-18
Rework hooks API internals

jwerle:
This PR also introduces new tests to cover the API which should look like:

```js
import hooks from 'socket:hooks'
hooks.onLoad(() => {
  // called when the runtime context has loaded,
  // but not ready or initialized
})

hooks.onReady(() => {
  // called when the runtime is ready, but not initialized
})

hooks.onInit(() => {
  // called when the runtime is initialized
})

hooks.onError((event) => {
  // called when 'error', 'messageerror', and 'unhandledrejection' error
  // events are dispatched on the global object
})

hooks.onData((event) => {
  // called when 'data' events are dispatched on the global object
})

hooks.onMessage((event) => {
  // called when 'message' events are dispatched on the global object
})

hooks.onOnline((event) => {
  // called when 'online' events are dispatched on the global object
 })

hooks.onOffline((event) => {
  // called when 'offline' events are dispatched on the global object
})
 ```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ jobs:
     - name: Check docs and TS definitions
       run: |
         npm run gen
+        git diff --quiet
 
   linux:
     name: Linux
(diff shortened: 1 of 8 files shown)


chicoxyzzy (approved):
Please run `npm run gen`

Actually `Lint` job should fail because docs and TS type definitions have changed. This was an intentional check to not to forget to push updated files.

Can you revert https://github.com/socketsupply/socket/commit/edf47d0c1d15a4ea44c4508c015597064c704493 ?

PR #457, opened 2023-05-24
chore(CI): print actionable error when docs or TS definitions are outdated

chicoxyzzy:


.github/workflows/ci.yml
```diff
@@ -23,6 +23,10 @@ jobs:
       run: |
         npm run gen
         git diff --quiet
+        if [$? -eq 1]; then
+          echo "Documentation or TypeScript definitions need update. Please do \`npm run gen\`."
```
mribbons, line 27: How about
```diff
+          echo "Documentation or TypeScript definitions are out of date. Please run \`npm run gen\`. to resolve"
```
Also, do the backticks really need to be escaped here?

mribbons (changes requested):
Note that `npm run gen` doesn't work on windows, some of the slashes end up the wrong way, eg:
```
### [`heartbeat()`](https://github.com/socketsupply/socket/blob/master/npm\packages\@socketsupply\socket-node\index.js#L198)
```

Can we also resolve that issue as part of this PR?

chicoxyzzy:
I don't have a Windows machine with me to test slashes in a docs generation script. Can we do it in another PR?

PR #458, opened 2023-05-24
IPC primordials cleanup

jwerle:


api/ipc.js
```diff
@@ -1409,6 +1409,8 @@ export function createBinding (domain, ctx) {
  * @ignore
  */
 export const primordials = sendSync('platform.primordials')?.data || {}
+Object.seal(primordials)
+Object.freeze(primordials)
```
chicoxyzzy, line 1413: do we need `seal` here?

chicoxyzzy (approved):

PR #461, opened 2023-05-25
SSC::Process: fix high cpu in wait()

mribbons:


jwerle (approved):

src/process/process.hh
```diff
@@ -185,7 +185,7 @@ namespace SSC {
 
     int wait () {
       do {
-        std::this_thread::yield();
+        msleep(256);
```
chicoxyzzy, line 188: maybe we can use something similar to this?

https://en.cppreference.com/w/cpp/thread/yield

chicoxyzzy: There's https://en.cppreference.com/w/cpp/thread/sleep_for which looks like what we need

jwerle: That is exactly what `msleep()` does :)

https://github.com/socketsupply/socket/blob/master/src/common.hh#L970-L972

chicoxyzzy: oh great!

chicoxyzzy (approved):

chicoxyzzy (approved):

src/process/process.hh
mribbons: @chicoxyzzy The `yield()` + loop approach might be the right approach for `msleep()`, but it's worth noting that `yield()` alone wasn't working here in terms of lower CPU usage. `msleep()` does call `yield()`, but it also calls `sleep()`.

PR #468, opened 2023-05-26
Enable inspectable webview in debug mode

jwerle:


heapwolf (approved):

src/ios/main.mm
```diff
@@ -270,6 +266,12 @@ -           (BOOL) application: (UIApplication*) application
   NSURL* url;
   int port = getDevPort();
 
+  if (@available(iOS 16.4, *)) {
```
chicoxyzzy, line 269: we can add tvOS too for future 😎

chicoxyzzy (approved):

src/ios/main.mm
jwerle: haha yeah we can come back to that - probably will want to have a `src/tvos/main.mm` file

PR #477, opened 2023-06-01
Android app start fix

mribbons:
cli: android: fail immediately if adb fails to install to device

Don't wait for emulator to boot.

+ improve error messages.

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -826,6 +826,9 @@ void runIOSSimulator (const Path& path, Map& settings) {
 struct AndroidCliState {
   StringStream adb;
   SSC::String androidHome;
+  // android, android-emulator
+  SSC::String targetPlatform;
+  // android-33 or current platform number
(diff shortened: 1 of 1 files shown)


lamiazar (approved):

jwerle (approved):

src/cli/cli.cc
```diff
@@ -2110,6 +2127,7 @@ int main (const int argc, const char* argv[]) {
     auto cnt = 0;
 
     AndroidCliState androidState;
+    androidState.targetPlatform = targetPlatform;
```
chicoxyzzy, line 2130: you may need to do the same for `ssc run` command

chicoxyzzy: maybe we can move the `androidState` fulfilment to the `run` function?

chicoxyzzy: > Just to confirm, does run get called by build?

yes

PR #486, opened 2023-06-08
fix(extension): debug on/off support for win32

mribbons:


jwerle:
CI failing

mribbons:
I think it's good now @jwerle, we're showing normal CI status.

jwerle (approved):

PR #492, opened 2023-06-14
Download 'sqlite3' in test build script

jwerle:


test/scripts/init.sh
chicoxyzzy, line 1: does it works on Windows?

jwerle: it does now!

chicoxyzzy (approved):

PR #506, opened 2023-07-22
fix(cli): run build script after copying

chicoxyzzy:


chicoxyzzy:
actually we need to keep the `[build] script` before `[build] copy` and add a separate step, something like `[build] post_copy_script`

chicoxyzzy:
I've changed my mind. Most of time we want to modify files after we copy them. For the 3rd party integrations it's usually a web app build and then build with ssc (which also does a files copying): `npm run build && ssc build -r`

I think we should merge this one

jwerle (approved):

PR #533, opened 2023-08-28
Support Socket bundle URI scheme

jwerle:
With these changes to how URIs work wrt to bundle identifiers routes will now work like below and by consequence you can import an ESM module from you app "namespace" (bundle identifier):

```js
import module from 'socket:com.beepboop/path/to/my/module'
```

and route sugar will work like:

- `socket://com.beepboop/ -> /index.html`
- `socket://com.beepboop/about -> /about/index.html`
- `socket://com.beepboop/pages/ -> /pages/index.html`
- `socket://com.beepboop/assets/video.mp4 -> /assets/video.mp4`

If a pathname has an extension, then it is explicit and no extension sugar will be applied
relative links work well in anchors too: `<a href="../">../</a>` 

Apps will be served from `socket://<bundle_identifier>/`, which is controlled by the developer
this should allow for features only available to domains and allow you to further control access with CSP
oauth redirects in the runtime webview can navigate to these URIs allowing existing auth to work 

The hope this allows projects like redwood and nextjs to "just work" with static builds as they typically expect a domain and a root / path from that domain, which just didn't work well with file:
dropping file: in favor of scoped assets to your app bundle removes unnecessarily exposing the file path to assets (file:///users/user/code/project/app/build/mac/my-app.app/Contents/Resources/index.html becomes socket://com.my-app/index.html)

this change brings back `ServiceWorker` and `Atomics` too


bcomnes:
Here is what I recommend:

- `socket://com.beepboop/` -> `/index.html` Trailing slash, directly serve the `index.html`
- ~~socket://com.beepboop/about -> /about.html~~ This is a non-standard index redirect. Perhaps we add support for [pillarjs/send](https://github.com/pillarjs/send) style index rule options at a later date, but the fundamental pattern we are implementing here is index.html resolution that is common across nginx, apache, and various node static libs. I think it would be a mistake to ship this by default.
- `socket://com.beepboop/pages/` -> `/pages/index.html` Directly serve the index.html from `/pages/`.
- `socket://com.beepboop/pages` -> `/pages/` 301 redirect to `/pages/` (trailing slash) if an `index.html` exists, otherwise 404. 
- `socket://com.beepboop/assets/video.mp4` -> `/assets/video.mp4` Directly served exact file match
- `socket://com.beepboop/page/index.html` -> `/page/index.html` Directly navigated html file serve up the same way

I think we should avoid adding any bespoke behavior here. Serving index.html files consistently with every other static HTML server by default should be the goal. If people want sugar like `about.html` -> `about/` perhaps we add support for custom index resolution rules, but let's not solve any new problems here. This is a well understood and implemented pattern across many implementations.

test/src/diagnostics/window.js
```diff
@@ -44,7 +44,7 @@ test('diagnostics - window - metrics', async (t) => {
       .then((response) => response.text())
       .then((text) => t.ok(typeof text === 'string', 'response.text()'))
       .then(() => t.ok(diagnosticsReceived, 'fetch diagnostics received'))
-      .catch((err) => t.ifError(err))
+      .catch((err) => t.ifError(err || true))
```
chicoxyzzy, line 47: what is this `|| true` for?

api/fetch/fetch.js
chicoxyzzy, line 1: do we need it to resolve the cookies issue?

package.json
```diff
@@ -23,7 +23,10 @@
     "acorn-walk": "8.2.0",
     "esbuild": "^0.17.19",
     "standard": "^17.1.0",
-    "typescript": "^5.1.3"
+    "typescript": "^5.1.3",
+    "urlpattern-polyfill": "^9.0.0",
+    "whatwg-fetch": "^3.6.17",
+    "whatwg-url": "^13.0.0"
```
chicoxyzzy, line 29: These modules are both in the `node_modules` and added as a vendor modules to the git. Was it intended?

api/fetch/fetch.js
bcomnes: We need to resolve the cookies issue if socket apps are to talk to hosted APIs that use cookies, but I think it's independent of this.

chicoxyzzy: @jwerle what's the purpose of the polyfill?

package.json
jwerle: yes so we know what the version of the node modules we vendor

chicoxyzzy: how do we sync them? do we have some script that copies them from `node_modules`?

test/src/diagnostics/window.js
jwerle: oops left over

package.json
jwerle: no script, but we could add one like:

```sh
esbuild --bundle --format=esm node_modules/whatwg-url/index.js > api/url/url/url.js
esbuild --bundle --format=esm node_modules/whatwg-fetch/index.js > api/fetch/fetch.js
```

jwerle:
> Here is what I recommend:
> 
> * `socket://com.beepboop/` -> `/index.html` Trailing slash, directly serve the `index.html`
> * ~socket://com.beepboop/about -> /about.html~ This is a non-standard index redirect. Perhaps we add support for [pillarjs/send](https://github.com/pillarjs/send) style index rule options at a later date, but the fundamental pattern we are implementing here is index.html resolution that is common across nginx, apache, and various node static libs. I think it would be a mistake to ship this by default.
> * `socket://com.beepboop/pages/` -> `/pages/index.html` Directly serve the index.html from `/pages/`.
> * `socket://com.beepboop/pages` -> `/pages/` 301 redirect to `/pages/` (trailing slash) if an `index.html` exists, otherwise 404.
> * `socket://com.beepboop/assets/video.mp4` -> `/assets/video.mp4` Directly served exact file match
> * `socket://com.beepboop/page/index.html` -> `/page/index.html` Directly navigated html file serve up the same way
> 
> I think we should avoid adding any bespoke behavior here. Serving index.html files consistently with every other static HTML server by default should be the goal. If people want sugar like `about.html` -> `about/` perhaps we add support for custom index resolution rules, but let's not solve any new problems here. This is a well understood and implemented pattern across many implementations.

done!

.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
   linux:
     name: Linux
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest-m
```
chicoxyzzy, line 29: we fixed this before to prevent unexpected updates which can break stuff

api/internal/init.js
```diff
@@ -14,10 +14,15 @@ console.assert(
...
 import { IllegalConstructor, InvertedPromise } from '../util.js'
 import { Event, CustomEvent, ErrorEvent } from '../events.js'
+import applyPolyfills from './polyfills.js'
+import location from '../location.js'
+import { URL } from '../url.js'
 
 const RUNTIME_INIT_EVENT_NAME = '__runtime_init__'
 const GlobalWorker = globalThis.Worker || class Worker extends EventTarget {}
 
+applyPolyfills()
```
getify, line 24: might be a NIT here, but... it may be confusing to refer to these as polyfills. that term is generally taken to mean "defines some behavior if the environment is missing that behavior natively". here, though, we're paving over the built-in ones, on purpose.

I just wouldn't want people to be confused thinking that these polyfills were conditional (and thus likely not being used in most envs) when in fact they'll be forced in all envs.

getify (approved):
LGTM

bcomnes (approved):
Sweet

PR #540, opened 2023-08-31
Fix 176

### LINKED ISSUE #176: windows - fix CI output has double line feeds

jwerle:
Fixes #176

### DIFF
--- a/src/common.hh
+++ b/src/common.hh
@@ -655,12 +655,12 @@ namespace SSC {
     auto& stream = isError ? std::cerr : std::cout;
     stream << str;
 
+  #if defined(_WIN32)
     if (IN_GITHUB_ACTIONS_CI) {
-    #if defined(_WIN32)
       notifyCli();
       return;
-    #endif
     }
+  #endif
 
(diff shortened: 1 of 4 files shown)


src/common.hh
```diff
@@ -686,7 +686,18 @@ namespace SSC {
...
+    static const auto IN_GITHUB_ACTIONS_CI = getEnv("GITHUB_ACTIONS_CI").size() == 0;
+    auto& stream = isError ? std::cerr : std::cout;
+    stream << str;
+
+    if (IN_GITHUB_ACTIONS_CI) {
+    #if defined(_WIN32)
+      notifyCli();
+      return;
+    #endif
+    }
```
chicoxyzzy, line 663: I think that `if (IN_GITHUB_ACTIONS_CI) {` should be inside of the `#if` directive

chicoxyzzy: but also I don't understand why skip `std::endl` on Windows

chicoxyzzy (approved):

PR #541, opened 2023-08-31
use 'hasEnv' for env value checks

### LINKED ISSUE #528: ENV vars defined in socket.ini always come in as '' when undefined
ENV vars that are configured in socket.ini always come in as a blank string when they are non set. They should resolve to `undefined` unless there is a reason not to.

jwerle:
Fixes #528


bcomnes (commented):
Hell yes, every detail counts.

bcomnes (approved):

Issue #556, opened 2023-09-06
Make `npm test` use the `--test` option


chicoxyzzy:
Currently we are explicitly adding test script to the `/test/index.html`

chicoxyzzy:
- we are checking if test file exists in the source directory; that's wrong, we should check it only in runtime
- we only should copy test files when `--test` is provided and `--prod` is not provided

PR #557, opened 2023-09-06
use the `--test` option for the test app

chicoxyzzy:
fixes https://github.com/socketsupply/socket/issues/556


jwerle (approved):
why drop `--prod` ?

chicoxyzzy:
> why drop `--prod` ?

I think you didn't refresh the page after my rebase. All `--prod` flags are there

PR #581, opened 2023-09-14
Fix internal cache reloading & control packet checks

heapwolf:
- internal cache was not being refilled properly
- intro control packets were not being cached

### DIFF
--- a/api/stream-relay/index.js
+++ b/api/stream-relay/index.js
@@ -503,7 +503,7 @@ export const wrap = dgram => {
     getState () {
       return {
         config: this.config,
-        data: [...this.cache.data.values()],
+        data: [...this.cache.data.entries()],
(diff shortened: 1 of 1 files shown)


api/stream-relay/index.js
```diff
@@ -503,7 +503,7 @@ export const wrap = dgram => {
     getState () {
       return {
         config: this.config,
-        data: [...this.cache.data.values()],
+        data: [...this.cache.data.entries()],
```
chicoxyzzy, line 506: why `[...this.cache.data.entries()]` and not just `this.cache.data.entries()`?

heapwolf: `map.entries()` returns `[Map Entries]` where as `[...map.entries()]` returns an array which means one less type of thing to deal with at serialization time.

PR #588, opened 2023-09-14
Handle various common mime types on Windows

jwerle:
Sometimes, a mime type cannot be determined from common extensions


chicoxyzzy (approved):
The list could become too big very soon. We can add webp, svg and maybe few other extensions

jwerle:
> The list could become too big very soon. We can add webp, svg and maybe few other extensions

yep see https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
😂

PR #594, opened 2023-09-14
FileSystemWatcher API

jwerle:
This PR introduces a file watching API for `ssc build` and in the runtime to emit `filedidchange` event. This API only works for desktop environments

 The `ssc build --watch` API allows watching files specified in the `socket.ini` config with:

```ini
[build]
copy = "..."

# or

[build]
copy_map = ""

# or

[build.copy_map]
src = dest

# or

[build.watch]
sources[] = "./directory" # can be a directory which will be recursive
sources[] = "./file.js" # or just a file
```

The files specified with `build.watch.sources` are the target files to watch that trigger the build phase. If they are not specified, then `build.copy` or `build.copy_map` files will be that target files to watch to trigger a build phase.

The runtime can be configured to watch for file changes in its resources directory with:

```ini
[webview]
watch = true
```

which will emit a `filedidchange` event on `globalThis` (`window`) on `event.detail.path` that points to the path relative to the resources directory of the application.


chicoxyzzy:
Also entries in the config template. We parse it with `npm run gen:doc` to create a config documentation which will be added to our website soon https://github.com/socketsupply/socketsupply.co/pull/259

chicoxyzzy (approved):

PR #597, opened 2023-09-15
Make `ssc build` fail early when `[meta] build_identifier` is not provided

chicoxyzzy:
TODO:
- [x] check that `build_identifier` is a valid string
- [x] check for other required fields (checks are there already)

Bonus:
- [x] more config docs

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -2413,9 +2413,19 @@ int main (const int argc, const char* argv[]) {
   createSubcommand("build", buildOptions, true, [&](Map optionsWithValue, std::unordered_set<String> optionsWithoutValue) -> void {
     auto isForExtensionOnly = settings["meta_type"] == "extension" || settings["build_type"] == "extension";
 
-    if (settings["meta_bundle_identifier"].size() == 0 !isForExtensionOnly) {
-      log("ERROR: [meta] bundle_identifier option is empty");
-      exit(1);
+    if (!isForExtensionOnly) {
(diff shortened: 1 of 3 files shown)


src/cli/cli.cc
```diff
@@ -2415,6 +2411,11 @@ int main (const int argc, const char* argv[]) {
   // Insert the elements of runOptions into buildOptions
   buildOptions.insert(buildOptions.end(), runOptions.begin(), runOptions.end());
   createSubcommand("build", buildOptions, true, [&](Map optionsWithValue, std::unordered_set<String> optionsWithoutValue) -> void {
+    if (settings["meta_bundle_identifier"].size() == 0) {
```
jwerle, line 2414: we should validate that the build is not for an extension

src/cli/templates.hh
```diff
@@ -1443,6 +1452,9 @@ maintainer = "Beep Boop Corp."
 ; The title of the app used in metadata files. This is NOT a window title. Can contain spaces and special characters. Defaults to name in a [build] section.
 title = "Beep Boop"
 
+; Builds an extension when set to "extension".
+type = "extension"
```
jwerle, line 1456: should this be here?

chicoxyzzy: Do you think we should comment it or it's better to remove it completely?

jwerle: Comment it out and set it to an empty string (default)

jwerle (approved):

PR #605, opened 2023-09-21
chore(docs): improve docs

chicoxyzzy:


chicoxyzzy:
@bcomnes what other modules from `socket:test` should I add?

api/test/dom-helpers.js
```diff
@@ -1,5 +1,21 @@
...
+ * Example usage:
+ * ```js
+ * import { test } from 'socket:test/dom-helpers.js'
+ * ```
+ *
+ */
 // Based on @socketsupply/test-dom and vhs-tape.
 
+const SECOND = 1000
+const defaultTimeout = 5 * SECOND
```
chicoxyzzy, line 17: #607

bcomnes (approved):
LGTM, but let's just keep dom-helpers out of the public docs for now.

jwerle (approved):

PR #616, opened 2023-09-26
Darwin Geolocation

jwerle:


api/internal/geolocation.js
```diff
@@ -0,0 +1,184 @@
...
+class Watcher {
+  constructor (identifier) {
+    this.identifier = identifier
+    this.listener = (event) => {
+      console.log(event)
+      if (event.detail?.params?.watch?.identifier === identifier) {
+        const position = createGeolocationPosition(event.detail.params)
+        console.log(position)
+      }
+    }
```
chicoxyzzy, line 19: does it just log to console? maybe we can improve the output to make it more descriptive?

jwerle: this should not be here!

chicoxyzzy (approved):

PR #617, opened 2023-09-26
Upd npm deps

chicoxyzzy:


package.json
```diff
@@ -19,11 +16,10 @@
   },
   "private": true,
   "devDependencies": {
-    "acorn": "8.8.2",
+    "acorn": "8.10.0",
     "acorn-walk": "8.2.0",
-    "esbuild": "^0.17.19",
     "standard": "^17.1.0",
-    "typescript": "~5.2.0",
+    "typescript": "5.2.2",
```
chicoxyzzy, line 22: I've pinned it so everyone has the same version when we build a release package

jwerle (approved):

package.json
```diff
@@ -19,11 +16,10 @@
   },
   "private": true,
   "devDependencies": {
-    "acorn": "8.8.2",
+    "acorn": "8.10.0",
     "acorn-walk": "8.2.0",
-    "esbuild": "^0.17.19",
```
chicoxyzzy, line 24: not used

bcomnes: Nice catch. Lots of tools come and go in the space, but last I checked https://github.com/webpro/knip is the best one at the moment for spotting these. Luckily we don't have to dig far to find these in our repos.

bcomnes (approved):

PR #621, opened 2023-09-26
Introduce missing fs javascript APIs

jwerle:


api/fs/index.js
```diff
@@ -601,35 +706,47 @@ export function stat (path, options, callback) {
...
 /**
  * @ignore
  */
-export function symlink (target, path, type, callback) {
-}
-/**
- * @ignore
- */
-export function truncate (path, length, callback) {
+export function symlink (src, dest, flags, callback) {
```
chicoxyzzy, line 713: the third parameter should be `type` (string | null)

api/fs/promises.js
```diff
@@ -294,38 +368,42 @@ export async function stat (path, options) {
 }
 
 /**
- * @TODO
  * @ignore
  */
-export async function symlink (target, path, type) {
-}
+export async function symlink (src, dest, flags) {
```
chicoxyzzy, line 371: third parameter should be of type string or null

chicoxyzzy (commented):
`readlink`, `realpath` and `rmdir` doesn't use the `options` parameter, but I think this is fine

chicoxyzzy (approved):

PR #632, opened 2023-10-02
SRP rc for socket 0.5

heapwolf:


jwerle (commented):
LGTM

@heapwolf can you run `npm run gen` to generate docs and types?

PR #634, opened 2023-10-03
Android geolocation with permissions

jwerle:


api/internal/geolocation.js
```diff
@@ -84,24 +104,70 @@ function createGeolocationPosition (data) {
...
+  getCurrentPosition: getPlatformFunction('getCurrentPosition'),
+  watchPosition: getPlatformFunction('watchPosition'),
+  clearWatch: getPlatformFunction('clearWatch')
 }
 
-export async function getCurrentPosition (onSuccess, onError, options = {}) {
+/**
+ * @param {function(GeolocationPosition)} onSuccess
+ * @param {onError(Error)} onError
+ * @param {?({ timeout?: number })} [options]
```
chicoxyzzy, line 130: not sure if that will parse correctly

jwerle: what should it be? 🙏🏼

chicoxyzzy:
```js
* @param {Object=} options
* @param {number=} options.timeout
```

src/core/preload.cc
```diff
@@ -127,22 +120,37 @@ namespace SSC {
...
+      "                                                                      \n"
+      "  try {                                                               \n"
+      "    const event = '__runtime_init__';                                 \n"
+      "    let onload = null                                                 \n"
+      "    Object.defineProperty(globalThis, 'onload', {                     \n"
+      "      get: () => onload,                                              \n"
+      "      set (value) {                                                   \n"
+      "        const opts = { once: true };                                  \n"
+      "        if (onload) {                                                 \n"
+      "          globalThis.removeEventListener(event, onload, opts);        \n"
```
chicoxyzzy, line 132: we probably don't need to remove it as it has `{ once: true }` already, but this is fine

src/window/apple.mm
```diff
@@ -681,55 +683,95 @@ -                       (void) webView: (WKWebView*) webView
...
+    } @catch (NSException *error) {
+      debug("Failed to set preference: 'storageAPIEnabled': %@", error);
     }
 
-    if (userConfig["permissions_allow_device_orientation"] == "false") {
-      [prefs setValue: @NO forKey: @"deviceOrientationEventEnabled"];
-    } else {
-      [prefs setValue: @YES forKey: @"deviceOrientationEventEnabled"];
+    @try {
+      if (userConfig["permissions_allow_device_orientation"] == "false") {
```
chicoxyzzy, line 707: do want want to add this to config?

PR #642, opened 2023-10-04
fix(cli): run user build script after package is scaffolded

aleclarson:
In my case, this is useful for manipulating the Info.plist file from my build script.


jwerle:
@aleclarson perhaps we can introduce a new life cycle here instead of moving the build phase script/copy around

Also, you could configure additions to the `Info.plist` with:

```ini
[mac.info.plist]
file = "file/to/mixin.plist"
# or
data = "data/to/mixin/into/Info.plist"
```

same for iOS:

```ini
[mac.info.plist]
file = "file/to/mixin.plist"
# or
data = "data/to/mixin/into/Info.plist"
```

we have not documented these features of the `socket.ini` yet

aleclarson:
> perhaps we can introduce a new life cycle here

what is the use case for running build script before scaffolding? nothing the build script could do would affect the scaffolding, so it's better to run it after so it can alter any of the project files before xcodebuild et al are applied

chicoxyzzy (approved):
I agree that build before scaffolding makes no sense.

LGTM

jwerle (approved):

Issue #511, opened 2023-07-28
Extensions: add and error when non-registered name is used in SAPI methods


chicoxyzzy:
Currently app will crash if SAPI methods are called with an unregistered name (i.e. `sapi_ipc_router_map(context, "UNKNOWN_EXTENSION.process.spawn", onProcSpawn, data)`). We should handle that case and add a descriptive error.

Other things to check:
- error in JS API `await extension.load('UNKNOWN-EXTENSION')`
- error in the `.process.spawn` part

PR #652, opened 2023-10-06
fix(cli): warn when extension sources are not found

aleclarson:
If a path in `build.extensions` does not exist or an extension's `socket.ini` has no source files declared, a warning will be logged.


jwerle (approved):

chicoxyzzy:
Oh, this is partially fixing https://github.com/socketsupply/socket/issues/511. Nice!

PR #656, opened 2023-10-06
fix: remove superfluous decodeURIComponent calls

aleclarson:
The IPC::Message constructor decodes the `value` and `seq` properties by default. If you pass true, the arguments will be decoded too.


jwerle (approved):
nice we've been meaning to getting around to this change

PR #658, opened 2023-10-06
fix: use --force with npm link

aleclarson:
When running `bin/publish-npm-modules.sh --link` repeatedly, the `npm link` can fail if you don't use `--force`.


jwerle (approved):

jwerle:
@chicoxyzzy maybe this will solve your linking issue?

chicoxyzzy (approved):
Yes!

PR #659, opened 2023-10-06
fix(ipc): stringify the value passed to `process.write`

aleclarson:
The value is expected by socket-node to be valid JSON:
  https://github.com/socketsupply/socket/blob/7bcf246bbaf96d93447df41e3c851d36ec3fd65b/npm/packages/@socketsupply/socket-node/index.js#L123

### DIFF
--- a/api/window.js
+++ b/api/window.js
@@ -288,7 +288,7 @@ export class ApplicationWindow {
       return await ipc.send('process.write', {
         index: this.#senderWindowIndex,
         event: options.event,
-        value: value ?? ''
+        value: value !== undefined ? JSON.stringify(value) : null
       })
     }
 


jwerle:
@chicoxyzzy can you take a look at this?
@aleclarson maybe we can type check before `JSON.stringify` in the event _raw strings_ are intended to write to the "backend"

aleclarson:
> maybe we can type check before `JSON.stringify` in the event _raw strings_ are intended to write to the "backend"

probably not worth the effort since `JSON.parse(JSON.stringify("raw string"))` works just as well (negligible perf penalty)

jwerle (approved):

PR #660, opened 2023-10-06
fix: send all process.stdout/process.stderr writes through ipc

aleclarson:
In some cases, user-created logs may be written directly to `process.stdout` and `process.stderr` so this PR intercepts those too.

Another PR incoming that will print the Node.js logs in the main process and send them to Console.app with os_log on macOS/iOS.

### DIFF
--- a/npm/packages/@socketsupply/socket-node/index.js
+++ b/npm/packages/@socketsupply/socket-node/index.js
@@ -16,6 +16,8 @@ class API {
 
   #buf = ''
   #emitter = new EventEmitter()
+  #writeStdout
+  #writeStderr
 
   constructor () {
     process.stdin.resume()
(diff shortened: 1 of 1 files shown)


npm/packages/@socketsupply/socket-node/index.js
```diff
@@ -85,10 +108,10 @@ class API {
 
     if (data.length > MAX_MESSAGE_KB) {
       const len = Math.ceil(data.length / 1024)
-      process.stderr.write(
+      this.#writeStderr(
```
jwerle, line 111: @chicoxyzzy thoughts on dropping this? and the `512` limit?

npm/packages/@socketsupply/socket-node/index.js
```diff
@@ -144,13 +167,13 @@ class API {
 
     if (s.length > MAX_MESSAGE_KB) {
       const len = Math.ceil(s.length / 1024)
-      process.stderr.write('WARNING: Sending large message to webview: ' + len + 'kb\n')
-      process.stderr.write('RAW MESSAGE: ' + s.slice(0, 512) + '...\n')
+      this.#writeStderr(
```
jwerle, line 170: @chicoxyzzy same here

chicoxyzzy: Yes, we can drop both. I don't know why we have such limit.

chicoxyzzy: @heapwolf maybe you know why?

jwerle (approved):

npm/packages/@socketsupply/socket-node/index.js
jwerle: It was left over from the legacy module. We can drop it. @aleclarson could you remove the limit and warning?

chicoxyzzy: Let’s leave it for now. I will check if it’s necessary with #666

chicoxyzzy (approved):

npm/packages/@socketsupply/socket-node/index.js
jwerle: SGTM

PR #665, opened 2023-10-07
feat(cli): allow subcommands, variables, and relative paths in extension compiler/linker flags

aleclarson:
Here's an example `socket.ini` for a native extension:
```ini
[meta]
name = "foo"
type = "extension"

[extension]
sources[] = ./src/extension.cc

[extension.compiler]
flags[] = -I./include

[extension.linker]
flags[] = $(pkg-config libsodium --libs)
flags[] = -L$PWD/node_modules/zeromq/build/libzmq/lib
flags[] = -lzmq
```

- Only values of `flags[]` are evaluated by this PR.
- Each value is searched for `./` and `../` paths to be resolved relative to the extension's root directory.
- Each value is searched for `$(…)` and replaces it with its execution result (trimmed and with `\n` replaced by spaces).
- Each value is searched for `$abc_123`-like substrings and replaces them with matching environment variables. Not sure if this is super necessary (the `exec` call when compiling/linking might already evaluate these?), but I wanted to make sure `$PWD` was replaced with the extension's root directory, instead of being transformed into the project root. I'm open to removing this though, if we feel like `$PWD` *should* point to the project root.

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -4599,16 +4599,69 @@ int main (const int argc, const char* argv[]) {
             }
 
             if (fs::exists(target)) {
+              target = fs::canonical(target);
+
               auto configFile = target / "socket.ini";
               auto config = parseINI(fs::exists(configFile) ? readFile(configFile) : "");
               settings["build_extensions_" + extension + "_path"] = target.string();
 
               for (const auto& entry : config) {
                 if (entry.first.starts_with("extension_sources")) {
-                  settings["build_extensions_" + extension] += entry.second;
+                  settings["build_extensions_" + extension] += fs::canonical(target / entry.second);
                 } else if (entry.first.starts_with("extension_")) {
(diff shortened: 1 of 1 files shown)


src/cli/cli.cc
```diff
@@ -4599,16 +4599,69 @@ int main (const int argc, const char* argv[]) {
...
                 } else if (entry.first.starts_with("extension_")) {
                   auto key = replace(entry.first, "extension_", extension + "_");
                   auto value = entry.second;
+                  if (key.ends_with("_flags")) {
+                    // Replace all $(…) with evaluated stdout.
+                    // E.g. $(pkg-config --libs --cflags libssl) => -lssl
+                    auto match = std::smatch{};
+                    while (std::regex_search(value, match, std::regex("\\$\\((.*?)\\)"))) {
+                      auto subcommand = match[1].str();
+                      log("Running subcommand: " + subcommand);
```
jwerle, line 4620: can this be behind a debug flag?

PR #667, opened 2023-10-07
fix: have `ipc.request` treat `params` argument as `params.value` if not a plain object

aleclarson:
- Introducing the `IPCSearchParams` class, which abstracts all the index/seq/value handling for send/sendSync/request/write to share

### DIFF
--- a/api/ipc.js
+++ b/api/ipc.js
@@ -1001,16 +1001,39 @@ export async function ready () {
   })
 }
 
+const { toString } = Object.prototype
+
+class IPCSearchParams extends URLSearchParams {
+  constructor (params, nonce) {
+    const index = params?.index ?? globalThis.__args?.index ?? 0
+    const seq = 'R' + nextSeq++
+
+    super({
+      value: params !== undefined && toString.call(params) !== '[object Object]'
+        ? { value: params }
+        : undefined,
+      ...params,
+      index,
+      seq,
+      nonce
+    })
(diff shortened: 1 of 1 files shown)


aleclarson:
Probably want to squash this one to keep the `chore(fix)` commits out of the master log

api/ipc.js
```diff
@@ -1299,10 +1289,9 @@ export async function request (command, params, options) {
 
   await ready()
 
-  const request = new globalThis.XMLHttpRequest()
   const signal = options?.signal
-  const index = globalThis?.__args?.index ?? 0
-  const seq = nextSeq++
+  const request = new globalThis.XMLHttpRequest()
+  const params = new IPCSearchParams(value, Date.now())
```
chicoxyzzy, line 1294: I think that we can set `noonce` in constructor

chicoxyzzy (commented):
It would be great to add e few tests, but we can do it later

api/ipc.js
aleclarson: `send` doesn't set nonce

jwerle (approved):

chicoxyzzy (approved):

chicoxyzzy (commented):
@aleclarson please run `npm run gen` to regenerate TypeScript type definitions

chicoxyzzy (approved):

PR #670, opened 2023-10-08
feat(extension): add `sapi_ipc_reply_with_error` function

aleclarson:
```c++
#include <socket/extension.h>

bool my_extension_function(sapi_ipc_result_t* result) {
  // Do some work...

  if (error_occurred) {
    const char* error_message = "An error occurred";
    return sapi_ipc_reply_with_error(result, error_message);
  } else {
    // Send the result...
    return sapi_ipc_reply(result);
  }
}
```


jwerle (approved):

jwerle:
nice this is super convenient

PR #674, opened 2023-10-09
feat(): create icon file on ssc init

chicoxyzzy:


src/cli/templates.hh
```diff
@@ -996,7 +996,7 @@ constexpr auto gXCodePlist = R"XML(<?xml version="1.0" encoding="UTF-8"?>
   <key>CFBundleIdentifier</key>
   <string>{{meta_bundle_identifier}}</string>
   <key>CFBundleIconFile</key>
-  <string>ui/icon.png</string>
+  <string>{{mac_icon}}</string>
```
jwerle, line 999: this should be `{{ios_icon}}`

jwerle (changes requested):

src/cli/templates.hh
chicoxyzzy, line 999: is this correct? @heapwolf

src/cli/templates.hh
```diff
@@ -1659,10 +1659,7 @@ headers = native-module1.hh
 ; cmd = "node backend/index.js"
 
 ; The icon to use for identifying your app on Windows.
-icon = ""
-
-; The icon to use for identifying your app on Windows.
-logo = "src/icons/icon.png"
```
chicoxyzzy, line 1665: we don't use it, seems like legacy code

src/cli/templates.hh
```diff
@@ -1624,17 +1624,14 @@ icon = "src/icon.png"
 
 [mac]
 
-; Mac App Store icon
-appstore_icon = "src/icons/icon.png"
-
```
chicoxyzzy, line 1629: not used

src/cli/templates.hh
```diff
@@ -1659,10 +1656,7 @@ headers = native-module1.hh
 ; cmd = "node backend/index.js"
 
 ; The icon to use for identifying your app on Windows.
-icon = ""
-
-; The icon to use for identifying your app on Windows.
-logo = "src/icons/icon.png"
```
chicoxyzzy, line 1665: legacy? changed all `win_logo` to be `win_icon`

heapwolf: win_logo was for the app package manifest, win_icon is the actual icon of the app, there may be a better name for it

chicoxyzzy: I brought `win_logo` back to not block this PR, we can change it's name later if necessary

chicoxyzzy: `[mac] appstore_icon` is not used anywhere so I removed it until we need it

jwerle (approved):

PR #676, opened 2023-10-10
Notification API

jwerle:
This pull request introduces support for the [`Notification API`](https://developer.mozilla.org/en-US/docs/Web/API/Notification). Developers will have permission to use this interface by default. It can be configured (opt-out) in the `[permissions]` section of the `socket.ini` file.

## Project Permissions

Permission to use this API can be configured in the `socket.ini` file:

```ini
[permissions]
allow_notifications = true # this is the default value
```

## User Permissions

Developers can request permission from the user to present/display notifications using the standard static [`Notification.requestPermission()`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/requestPermission_static) method.

```js
const status = await Notification.requestPermission() // can be 'granted', 'default', 'denied'
```

The current permission can be checked with the static [`Notification.permission`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/permission_static) property.

```js
Notification.permission // can be 'granted', 'default', 'denied'
```

### Permissions API (`navigator.permissions`)

Developers can use the [`Permissions API`](https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API) to `query()` for the current status of the `'notifications'` permission.

```js
const result = await navigator.permissions.query({ name: 'notifications' }}) 
console.log(result.state) // can be 'granted', 'prompt', 'denied'
```

There exists a non-standard `request()` API on the [`Permissions API`](https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API) at `navigator.permissions.request({ name: 'permission' })` that can be used.

```js
const result = await navigator.permissions.request({ name: 'notifications' }}) 
console.log(result.state) // can be 'granted', 'prompt', 'denied'
```

## Example

```js
const status = await navigator.permissions.query({ name: 'notifications' })
status.addEventListener('change', async () => {
  if (status.state === 'prompt') {
    await navigator.permissions.request({ name: 'notifications' })
  } else if (status.state === 'granted') {
    const notification = new Notification('New Notification', {
      title: 'New Message',
      body: 'New message from @jwerle',
      tag: '@jwerle'
    })

    let dismissed = true
    notification.addEventListener('click', () => {
      dismissed = false
      // handle notification click
    })

    notification.addEventListener('close', () => {
      if (!dismissed) {
        // handle closed notification, that was not dismissed
      }
    })
  }
})
```

## Platform Support

- [x] Android - supported by [`androidx.core.app.NotificationCompat`](https://developer.android.com/reference/kotlin/androidx/core/app/NotificationCompat)
- [x] iOS - supported by [`UNNotification`](https://developer.apple.com/documentation/usernotifications?language=objc)
- [x] Linux - supported natively, ingrates with runtime `socket:notification` and `socket:internal/permissions` APIs
- [x] macOS - supported by [`UNNotification`](https://developer.apple.com/documentation/usernotifications?language=objc)
- [ ] Windows - non-supported in regular release, yet. See [WebView2Feedback#308](https://github.com/MicrosoftEdge/WebView2Feedback/issues/308), [Release Notes for the WebView2 SDK
](https://learn.microsoft.com/en-us/microsoft-edge/webview2/release-notes?tabs=win32cpp#experimental-features), and [`interface ICoreWebView2Experimental22
`](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2experimental22?view=webview2-1.0.1988-prerelease&preserve-view=true#add_notificationreceived). When the interfaces provided in `ICoreWebView2Experimental22` move to `ICoreWebView2_22`, we will support them and then provide support for the `Notification` API in Windows. At the moment, using these APIs on Windows will lead to undefined behaviour.


src/cli/templates.hh
```diff
@@ -1637,11 +1827,10 @@ category = ""
 icon = ""
 
 ; TODO Signing guide: https://socketsupply.co/guides/#code-signing-certificates
```
chicoxyzzy, line 1829: Let's remove `TODO` here, so it doesn't appear in the docs

chicoxyzzy (approved):

bcomnes (approved):
I read through the whole thing and didn't see any standouts. Nice work, can't wait to try it out.

PR #677, opened 2023-10-10
fix(ipc): omit `value` parameter if value is undefined

aleclarson:
I've seen instances where `value=undefined` was included, which isn't a huge deal, but also not intended behavior.


jwerle:
just ran into this in #676 !

jwerle (approved):

PR #679, opened 2023-10-10
feat(extension): add `sapi_ipc_send_chunk` and `sapi_ipc_send_event`

aleclarson:
### ⚠️ Only MacOS and iOS are supported by this PR. I don't have time to implement the other platforms at the moment.

- Before calling sapi_ipc_send_chunk, you must first set the Transfer-Encoding header to `chunked` and then call sapi_ipc_reply.
- When using sapi_ipc_send_chunk, the frontend should use XMLHttpRequest. Currently, the Fetch API does not support response streaming in WebKit.
- Before calling sapi_ipc_send_event, you must first set the Content-Type header to `text/event-stream` and then call sapi_ipc_reply.
- When using sapi_ipc_send_event, the frontend should use EventSource.

### DIFF
--- a/include/socket/extension.h
+++ b/include/socket/extension.h
@@ -1097,6 +1097,51 @@ extern "C" {
   );
 
   /**
+   * Write a chunk to the HTTP response associated with the IPC result.
+   *
+   * ⚠️ You must have called `sapi_ipc_result_set_header` with a `Transfer-Encoding`
+   * header of `chunked` to use this function.
+   *
+   * ⚠️ You must call `sapi_ipc_reply` before calling this function.
+   *
+   * @param result      - An IPC request result
+   * @param chunk       - The chunk to write
+   * @param chunk_size  - The size of the chunk
+   * @param finished    - `true` if this is the last chunk to write
+   */
+  SOCKET_RUNTIME_EXTENSION_EXPORT
+  bool sapi_ipc_send_chunk (
+    sapi_ipc_result_t* result,
+    const char* chunk,
+    size_t chunk_size,
+    bool finished
+  );
+
+  /**
+   * Write an event to the HTTP response associated with the IPC result.
+   *
+   * ⚠️ You must have called `sapi_ipc_result_set_header` with a `Content-Type`
+   * header of `text/event-stream` to use this function.
+   *
+   * ⚠️ You must call `sapi_ipc_reply` before calling this function.
+   *
+   * ⚠️ The `name` and `data` arguments must be null-terminated strings. Either
+   * can be empty as long as it's null-terminated and the other is not empty.
+   *
+   * @param result   - An IPC request result
+   * @param name     - The event name
+   * @param data     - The event data
+   * @param finished - `true` if this is the last event to write
+   */
+  SOCKET_RUNTIME_EXTENSION_EXPORT
+  bool sapi_ipc_send_event (
+    sapi_ipc_result_t* result,
+    const char* name,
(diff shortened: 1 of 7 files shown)


include/socket/extension.h
```diff
@@ -1096,6 +1096,51 @@ extern "C" {
...
+  SOCKET_RUNTIME_EXTENSION_EXPORT
+  bool sapi_ipc_send_chunk (
+    sapi_ipc_result_t* result,
+    const char* chunk,
+    size_t chunk_size,
+    bool finished
+  );
+
+  /**
+   * Write an event to the HTTP response associated with the IPC result.
```
jwerle, line 1121: Can we include a note about this working on iOS/macOS only?

include/socket/extension.h
```diff
@@ -1096,6 +1096,51 @@ extern "C" {
     const sapi_json_any_t* json
   );
 
+  /**
+   * Write a chunk to the HTTP response associated with the IPC result.
```
jwerle, line 1100: Can we include a note about this working on iOS/macOS only?

jwerle (changes requested):
Left some comments. We'll need to set `isHTTP = true` in `src/window/win.cc` too. I wonder if we need a more general way to say where the source of the `IPC::Message` is from cc @chicoxyzzy

aleclarson:
👍

jwerle (approved):

chicoxyzzy (commented):
We have to check that this change doesn't break other platforms. Should we just allow CI for external PRs for now?

src/extension/ipc.cc
```diff
@@ -32,12 +32,7 @@ bool sapi_ipc_router_map (
     auto router,
     auto reply
   ) mutable {
-    auto msg = SSC::IPC::Message(
-      message.uri,
-      true, // decode parameter values AOT in `message.uri`
-      message.buffer.bytes,
-      message.buffer.size
-    );
+    auto msg = SSC::IPC::Message(message);
```
jwerle, line 35: We wanted an AOT decode so the `sapi_ipc_message_get()` calls get a decoded value

aleclarson: I've guaranteed that `Router::invoke` receives an `IPC::Message` with decoded values in other parts of this PR.

https://github.com/aleclarson/socket/blob/b9449443918685f6b2d15cca18ad62efd182ed06/src/ipc/bridge.cc#L2728-L2736

When `Router::invoke` is called with a `uri` string, values are always decoded (is this a problem?). When `Router::invoke` is called with a `IPC::Message` instead, it's assumed that the message has decoded values, which I've checked is true anywhere this overload is used as of now.

src/extension/ipc.cc
```diff
@@ -609,6 +654,24 @@ void sapi_ipc_result_set_header (
...
+              });
+    } else if (strcasecmp(name, "transfer-encoding") == 0 &&
+               strcasecmp(value, "chunked") == 0) {
+      result->post = SSC::Post();
+      result->post.chunk_stream =
+          std::make_shared<std::function<bool(const char*, size_t, bool)>>(
+              [result](const char* chunk, size_t chunk_size, bool finished) {
+                return false;
+              });
+    }
```
aleclarson, line 674: We could probably move this to `sapi_ipc_reply` later as a perf optimization.

chicoxyzzy (approved):

jwerle (approved):

PR #686, opened 2023-10-12
fix: setSystemMenu JS validation

aleclarson:
The validation logic in `application.js` was too strict, as the proper way to declare an unbound menu item (i.e. you must click the item, no key binding exists) is by omitting a `:` from the line.

### DIFF
--- a/api/application.js
+++ b/api/application.js
@@ -271,25 +271,48 @@ export async function setSystemMenu (o) {
   const frame = e.stack.split('\n')[2]
   const callerLineNo = frame.split(':').reverse()[1]
 
+  // Use this link to test the regex (https://regexr.com/7lhqe)
+  const validLineRegex = /^(?:([^:]+)|(.+)[:][ ]*((?:[+\w]+(?:[ ]+|[ ]*$))*)(.*))$/m
+  const validModifiers = /^(Alt|CommandOrControl|Control|Meta)$/
+
   for (let i = 0; i < lines.length; i++) {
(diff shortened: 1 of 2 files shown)


api/application.js
```diff
@@ -271,25 +271,48 @@ export async function setSystemMenu (o) {
   const frame = e.stack.split('\n')[2]
   const callerLineNo = frame.split(':').reverse()[1]
 
+  // Use this link to test the regex (https://regexr.com/7lhqe)
+  const validLineRegex = /^(?:([^:]+)|(.+)[:][ ]*((?:[+\w]+(?:[ ]+|[ ]*$))*)(.*))$/m
```
aleclarson, line 275: https://regexr.com/7lhqe

api/application.js
```diff
@@ -297,6 +297,8 @@ export async function setSystemMenu (o) {
       const binding = match[3]
       if (label.length === 0) {
         err = 'Missing label'
+      } else if (label.includes(':')) {
+        err = 'Invalid label contains ":"'
```
aleclarson, line 301: I think it'd be preferable to allow labels with `:` in the future by changing the native code to search for the last `:` instead of splitting by all `:` characters.

For now, an error is thrown.

chicoxyzzy (commented):
LGTM

please run `npm run gen` to update the API docs (so they point to correct line nubers)

chicoxyzzy (approved):

jwerle (approved):

PR #687, opened 2023-10-12
Breakout core runtime files

jwerle:
This PR breaks out core runtime concerns into their own files, moving inline implementations out of the headers, and subsequently improving the build time feedback loop with less changes occurring in the head files to trigger a complete rebuild. This PR renames various functions/macros for improved naming. It also removes the need for `common.hh` while also improving the header include surface for `init.cc`, meaning less downstream dependency requirements.


src/cli/cli.cc
```diff
@@ -4751,7 +4797,7 @@ int main (const int argc, const char* argv[]) {
 
       // Check for gradle in pwd. Don't fail, this is just for support.
       if (!fs::exists(String("gradlew") + (platform.win ? ".bat" : ""))) {
-        log("WARN: gradlew script not in pwd: " + fs::current_path().string());
+        log("WARNING: gradlew script not in pwd: " + fs::current_path().string());
```
chicoxyzzy, line 4800: we can introduce `warn` and `error` functions (later) to prefix automatically

jwerle: 👍🏼

bcomnes (approved):
Big organizational improvement.

api/notification.js
```diff
@@ -876,7 +876,7 @@ export class Notification extends EventTarget {
 }
 
 hooks.onReady(() => {
-  if (os.host() === 'iphone-simulator' || os.host() === 'android-emulator') {
+  if (os.host() === 'iphonesimulator' || os.host() === 'android-emulator') {
```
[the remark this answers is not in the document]
jwerle: it is the same value used with `xcrun -sdk` - we could include a `-` if you think appropriate

PR #688, opened 2023-10-12
fix(types): extend Window type with built-in events

aleclarson:


chicoxyzzy (commented):
were those added with `npm run gen` or manually?

aleclarson:
Ah good point!

chicoxyzzy:
Yeah, we generate type defs with `npm run gen`, so manually added stuff will be overwritten

aleclarson:
I don't think it's an issue. I'm seeing the types in the `build/npm` package, no problem.

jwerle (approved):

PR #692, opened 2023-10-14
Some fixes to `send_event` and `send_chunk` extension APIs

aleclarson:
- Correctly retain the context when:
  - Transfer-Encoding header is set to `chunked`
  - Content-Type header is set to `text/event-stream`
- Add debug log for incorrect use (failure to set expected header)

### DIFF
--- a/src/extension/ipc.cc
+++ b/src/extension/ipc.cc
@@ -167,7 +167,10 @@ bool sapi_ipc_send_chunk (
   }
   bool success = (*send_chunk_ptr)(chunk, chunk_size, finished);
   if (finished) {
-    sapi_context_release(result->context);
+    auto context = result->context;
+    if (context->release()) {
+      delete context;
+    }
   }
   return success;
 }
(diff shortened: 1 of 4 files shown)


jwerle (approved):

src/extension/ipc.cc
```diff
@@ -137,8 +137,7 @@ bool sapi_ipc_reply (const sapi_ipc_result_t* result) {
   }
 
   // if retained, then then caller must eventually call `sapi_context_release()`
-  if (!context->retained) {
-    context->release();
+  if (context->release()) {
```
aleclarson, line 140: Ah wait, this doesn't seem right

aleclarson: For this to work, we have to allocate the route context for each message. See my latest commit.

jwerle:
@aleclarson LGTM - is this ready?

aleclarson:
👍

PR #701, opened 2023-10-16
fix(mac): avoid "This task has already been stopped" exceptions

aleclarson:
**Status:** Seems to work

### Context
macOS does not allow `didReceiveResponse` or `didReceiveData` calls if a URL scheme task is cancelled.

### DIFF
--- a/src/ipc/bridge.cc
+++ b/src/ipc/bridge.cc
@@ -2333,7 +2333,18 @@ static void registerSchemeHandler (Router *router) {
 
 #if defined(__APPLE__)
 @implementation SSCIPCSchemeHandler
-- (void) webView: (SSCBridgedWebView*) webview stopURLSchemeTask: (Task) task {}
+{
+  NSMutableSet<Task>* _tasks;
+}
+- (instancetype) init {
+  if (self = [super init]) {
+    _tasks = [NSMutableSet set];
+  }
+  return self;
+}
+- (void) webView: (SSCBridgedWebView*) webview stopURLSchemeTask: (Task) task {
+  [_tasks removeObject: task];
+}
 - (void) webView: (SSCBridgedWebView*) webview startURLSchemeTask: (Task) task {
   static auto userConfig = SSC::getUserConfig();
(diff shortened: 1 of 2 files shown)


src/ipc/bridge.cc
```diff
@@ -2601,7 +2612,14 @@ static void registerSchemeHandler (Router *router) {
     body = (char *) data;
   }
 
+  [_tasks addObject: task];
```
aleclarson, line 2615: Only `router->invoke` tasks are added, as the other possibilities are synchronous if I'm not mistaken.

src/ipc/bridge.cc
```diff
@@ -2616,6 +2634,9 @@ static void registerSchemeHandler (Router *router) {
     if (result.post.event_stream != nullptr) {
       *result.post.event_stream = [task](const char* name, const char* data,
                                          bool finished) {
+        if (![_tasks containsObject: task]) {
+          return false;
```
aleclarson, line 2638: Callers of `sapi_ipc_send_event` or send_chunk should stop their work if `false` is returned.

src/ipc/bridge.cc
```diff
@@ -2601,7 +2612,14 @@ static void registerSchemeHandler (Router *router) {
     body = (char *) data;
   }
 
+  [_tasks addObject: task];
+
   auto invoked = self.router->invoke(message, body, bufsize, [=](Result result) {
+    // @TODO Communicate task cancellation to the route, so it can cancel its work.
```
aleclarson, line 2618: In the future, we'll want a way for route invocations to register a callback for task cancellation.

jwerle (approved):

aleclarson:
I've come to the conclusion that `sapi_ipc_send_event`/`send_chunk` callers should make sure they're on the main thread first (using `sapi_context_dispatch` if needed), instead of trying to use atomics to prevent race conditions (which I seemingly failed at anyway).

jwerle (approved):

PR #704, opened 2023-10-17
fix(socket-node): avoid resolving `send` promise with Error object

aleclarson:
It seemed odd to need to handle errors as a promise result, rather than through a catch callback.


jwerle:
yeah this is an internal pattern that should not have leaked into user space

jwerle (approved):

chicoxyzzy (approved):

PR #709, opened 2023-10-18
fix(socket-node): include stack trace for uncaught exceptions

aleclarson:


npm/packages/@socketsupply/socket-node/index.js
```diff
@@ -28,7 +28,7 @@ class API {
       this.#write(`ipc://process.exit?value=${exitCode}`)
     })
     process.on('uncaughtException', (err) => {
-      this.#write(`ipc://stderr?value=${encodeURIComponent(String(err))}`)
+      this.#write(`ipc://stderr?value=${encodeURIComponent(err.stack || String(err))}`)
```
aleclarson, line 31: Don't know when `stack` would be undefined, but TypeScript says it might be 🤷

jwerle (approved):

PR #712, opened 2023-10-18
Persistent file permissions on macOS

bcomnes:
WIP persistent file permissions on macOS.


jwerle:
closing as we took a different approach

PR #724, opened 2023-10-26
fix(src/ipc/bridge.cc): add missing ios branch for base path in router resolution

jwerle:


chicoxyzzy (approved):
The fix works on my laptop

Issue #714, opened 2023-10-19
Implement mkdir recursive


bcomnes:
We need to implement the `recursive` flag on `mkdir` because its extremely handy.

PR #729, opened 2023-10-26
Add a recursive option to mkdir

bcomnes:
This adds a recursive option to mkdir. 

Closes https://github.com/socketsupply/socket/issues/714

### DIFF
--- a/api/fs/index.js
+++ b/api/fs/index.js
@@ -396,7 +396,7 @@ export function mkdir (path, options, callback) {
   }
 
   const mode = options.mode || 0o777
-  const recursive = options.recursive ?? false
+  const recursive = Boolean(options.recursive)
 
   if (typeof mode !== 'number') {
     throw new TypeError('mode must be a number.')
--- a/api/fs/index.js
+++ b/api/fs/index.js
@@ -396,7 +396,7 @@ export function mkdir (path, options, callback) {
   }
 
   const mode = options.mode || 0o777
-  const recursive = options.recursive === true
+  const recursive = options.recursive === false
 
   if (typeof mode !== 'number') {
     throw new TypeError('mode must be a number.')
--- a/api/fs/index.js
+++ b/api/fs/index.js
@@ -396,7 +396,7 @@ export function mkdir (path, options, callback) {
   }
 
   const mode = options.mode || 0o777
-  const recursive = Boolean(options.recursive)
+  const recursive = Boolean(options.recursive) // default to false
 
   if (typeof mode !== 'number') {
     throw new TypeError('mode must be a number.')
--- a/api/fs/index.js
+++ b/api/fs/index.js
@@ -396,7 +396,7 @@ export function mkdir (path, options, callback) {
   }
 
   const mode = options.mode || 0o777
-  const recursive = options.recursive === false
+  const recursive = options.recursive ?? false
 
   if (typeof mode !== 'number') {
     throw new TypeError('mode must be a number.')
--- a/src/core/core.hh
+++ b/src/core/core.hh
@@ -510,7 +510,8 @@ namespace SSC {
             const String seq,
             const String path,
             int mode,
-            Module::Callback cb
+            Module::Callback cb,
+            bool recursive = false
           );
           void readlink (
             const String seq,
--- a/src/core/core.hh
+++ b/src/core/core.hh
@@ -510,8 +510,8 @@ namespace SSC {
             const String seq,
             const String path,
             int mode,
-            Module::Callback cb,
-            bool recursive = false
+            bool recursive,
+            Module::Callback cb
           );
           void readlink (
             const String seq,
--- a/src/core/fs.cc
+++ b/src/core/fs.cc
@@ -1841,53 +1841,85 @@ namespace SSC {
     });
   }
 
-  void Core::FS::mkdir (
+  static void mkdirRecursive(
+    const String& path,
+    int mode,
+    uv_loop_t* loop,
+    std::function<void(int)> callback
+  ) {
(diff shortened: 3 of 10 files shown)


api/fs/index.js
```diff
@@ -379,7 +379,7 @@ export function mkdir (path, options, callback) {
   }
 
   const mode = options.mode || 0o777
-  const recursive = options.recursive === true
+  const recursive = options.recursive === false
```
chicoxyzzy, line 382: Oh that's confusing. Is it like that in Node?

bcomnes: Node defaults to false because it didn't used to have that behavior (hence the use of the mkdirp package everywhere in the ecosystem).

bcomnes: We could default to true, but then we would be different than node.  ¯\_(ツ)_/¯

chicoxyzzy: Ah, this is not default :) There's no default here. This check only makes `recursive` equal `true` if the option that passed is a boolean `true` (not object or string `"true"`, etc.).

bcomnes: Oh yeah, I totally read that wrong, fixing.

src/core/core.hh
```diff
@@ -506,7 +506,8 @@ namespace SSC {
             const String seq,
             const String path,
             int mode,
-            Module::Callback cb
+            Module::Callback cb,
+            bool recursive = false
```
jwerle, line 510: Could we put those before `cb` and after `mode`?

src/core/fs.cc
```diff
@@ -1714,53 +1714,85 @@ namespace SSC {
...
+    uv_fs_stat(loop, &stat_req, filename, nullptr);
+
+    if (stat_req.result == 0) {
+        uv_fs_req_cleanup(&stat_req);
+        callback(0);
+        return;
+    }
+    uv_fs_req_cleanup(&stat_req);
+
+    size_t last_slash = path.rfind('/');
```
jwerle, line 1735: we'll want to check for `\\` on Windows

src/ipc/bridge.cc
```diff
@@ -736,14 +736,15 @@ static void initRouterTable (Router *router) {
...
+    /**
+   * Creates a directory at `path` with an optional mode and an optional recursive flag.
    * @param path
    * @param mode
+   * @param recursive
    * @see mkdir(2)
    */
   router->map("fs.mkdir", [](auto message, auto router, auto reply) {
-    auto err = validateMessageParameters(message, {"path", "mode"});
+    auto err = validateMessageParameters(message, {"path", "mode", "recursive"});  // Added "recursive" here
```
jwerle, line 747: Adding `"recursive"` here makes this a required argument. Is that needed?

src/core/core.hh
bcomnes: I tried that but clang got mad about having defaulted arguments come before the callback.

src/ipc/bridge.cc
bcomnes: I want to make it an optional field in the options object in the JS api, but it looks like the JS side forces a value before calling the c++ bridge. I could be reading it wrong too. Happy to change anything.

src/core/core.hh
jwerle: lets just avoid the default value here and pass the recursive option where ever this function is used

src/core/fs.cc
bcomnes: Whats the best pattern for doing that? Something like `path.sep` from node? A C++ macro?

jwerle: something like
```c++
#if defined(_WIN32)
size_t last_slash = path.find("\\");
#else
size_t last_slash = path.find("/");
#endif
```

src/ipc/bridge.cc
```diff
@@ -736,14 +736,15 @@ static void initRouterTable (Router *router) {
     );
   });
 
-  /**
-   * Creates a directory at `path` with an optional mode.
+    /**
```
chicoxyzzy, line 739: I think that there's an extra space here

src/ipc/bridge.cc
```diff
@@ -736,10 +736,11 @@ static void initRouterTable (Router *router) {
     );
   });
 
-  /**
-   * Creates a directory at `path` with an optional mode.
+ /**
```
chicoxyzzy, line 739: now there is one more space needed :)
it should look like this (the first `*` in `/**` should be on the same position as the following `*`s)
```
/**
 *
 */
```

bcomnes: Fixed.

src/core/fs.cc
```diff
@@ -1,4 +1,6 @@
 #include "core.hh"
+#include "uv.h"
+#include <sys/_types/_s_ifmt.h>
```
bcomnes, line 3: @jwerle I forget, do we still need this?

bcomnes:
Ok, I think this should be good to go.

src/core/fs.cc
jwerle: nope we can remove that

bcomnes: Both of them?

bcomnes: It's both of them.

chicoxyzzy (approved):
LGTM 👍 
tests pass on my computer

bcomnes:
Looking at CI failures a little closer.

jwerle (approved):

bcomnes:
The recursive tests are passing now. The fs.writeFile isn't working on master so I'll merge this now.

PR #731, opened 2023-10-27
Fix test ordering

bcomnes:
When you import a tests, they can run out of expected order. This guarantees that tests run in order.

Short term fix until we have built in test globbig/loading.


jwerle:
`await import()` will break tests for `widow.onload` as now ESM is async.

bcomnes:
Arg you're right.

PR #734, opened 2023-10-29
allow higher limits on open files ios/macos

heapwolf:


src/desktop/main.cc
```diff
@@ -54,6 +54,14 @@ SSC::String getNavigationError (const String &cwd, const String &value) {
 // which on windows is hInstance, on mac and linux this is just an int.
 //
 MAIN {
+
```
jwerle, line 57: I don't think we need to do this here as we have the Plist setting. We're also losing control from the user `[meta] file_limit` setting

heapwolf: ah I thought that wasn't getting applied, but I ran the check and it looks like it is actually getting applied on desktop

jwerle (approved):

PR #736, opened 2023-10-30
Introduce 'fs.watch' and 'fs.Watcher'

### LINKED ISSUE #595: 'fs.watch()' API for Desktop
We should introduce a `fs.watch()` API based on the implementation in #594 that is similar to node's https://nodejs.org/api/fs.html#fswatchfilename-options-listener

jwerle:
This PR introduces the `fs.watch()` and `fs.Watcher` APIs. It is implemented to give as much parity as possible with node's implementation.

### Examples

`EventEmitter`:

```js
import process from 'socket:process'
import fs from 'socket:fs'

// watch resources directory
const watcher  = fs.watch(process.cwd())

watcher.on('change', (eventType, filename) => {
  console.log(eventType) // 'change' or 'rename'
  console.log(filename) // file name relative to `process.cwd()`
})
```

`AsyncIterator` (with `filename` encoding):

```js
import process from 'socket:process'
import fs from 'socket:fs/promises'

// output `filename` as `Buffer` instance
const events = fs.watch(process.cwd(), { encoding: 'buffer' })

for await (const event of events) {
  console.log(event) // { eventType: 'change' | 'rename' , filename: <Buffer ...> }
}
```

`AbortSignal`

```js
import process from 'socket:process'
import fs from 'socket:fs/promises'

const controller = new AbortController()
// output `filename` as `Buffer` instance
const events = fs.watch(process.cwd(), { signal: controller.signal })

// watch files for 5000 ms
setTimeout(() => controller.abort(), 5000)

for await (const event of events) {
  console.log(event) // { eventType: 'change' | 'rename' , filename: 'index.js' }
}
```

Closes #595

### DIFF
--- a/api/fs/dir.js
+++ b/api/fs/dir.js
@@ -61,11 +61,29 @@ export class Dir {
   }
 
   /**
+   * `true` if closed, otherwise `false`.
+   * @ignore
+   * @type {boolean}
+   */
+  get closed () {
+    return Boolean(this.handle?.closed)
+  }
+
+  /**
+   * `true` if closeing, otherwise `false`.
+   * @ignore
+   * @type {boolean}
+   */
+  get closing () {
+    return Boolean(this.handle?.closig)
+  }
+
+  /**
    * Closes container and underlying handle.
    * @param {object|function} options
    * @param {function=} callback
    */
-  async close (options, callback) {
+  async close (options = null, callback) {
     if (typeof options === 'function') {
       callback = options
       options = {}
--- a/api/fs/dir.js
+++ b/api/fs/dir.js
@@ -70,12 +70,12 @@ export class Dir {
   }
 
   /**
-   * `true` if closeing, otherwise `false`.
+   * `true` if closing, otherwise `false`.
    * @ignore
    * @type {boolean}
    */
   get closing () {
-    return Boolean(this.handle?.closig)
+    return Boolean(this.handle?.closing)
   }
 
   /**
--- a/api/fs/index.js
+++ b/api/fs/index.js
@@ -854,6 +854,7 @@ export function writeFile (path, data, options, callback) {
  * Watch for changes at `path` calling `callback`
  * @param {string}
  * @param {function|object=} [options]
+ * @param {string=} [options.encoding = 'utf8']
  * @param {?function} [callback]
  * @return {Watcher}
  */
--- a/api/fs/index.js
+++ b/api/fs/index.js
@@ -33,6 +33,7 @@ import { DirectoryHandle, FileHandle } from './handle.js'
 import { ReadStream, WriteStream } from './stream.js'
 import * as constants from './constants.js'
 import * as promises from './promises.js'
+import { Watcher } from './watcher.js'
 import { Stats } from './stats.js'
 import fds from './fds.js'
 
@@ -47,7 +48,7 @@ function defaultCallback (err) {
   if (err) throw err
 }
 
-async function visit (path, options, callback) {
+async function visit (path, options = null, callback) {
   if (typeof options === 'function') {
     callback = options
     options = {}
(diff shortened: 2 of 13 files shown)


api/fs/index.js
```diff
@@ -346,7 +360,10 @@ export function lchown (path, uid, gid, callback) {
 }
 
 /**
- * @ignore
+ * Creates a link to `dest` from `dest`.
```
chicoxyzzy, line 363: from `src` to `dest`?

api/fs/index.js
```diff
@@ -820,6 +850,24 @@ export function writeFile (path, data, options, callback) {
   })
 }
 
+/**
+ * Watch for changes at `path` calling `callback`
+ * @param {string}
+ * @param {function|object=} [options]
+ * @param {string=} [options.encoding = 'utf8']
```
chicoxyzzy, line 857: this one is not in the params list

api/fs/promises.js
```diff
@@ -455,4 +486,16 @@ export async function writeFile (path, data, options) {
...
 
+/**
+ * Watch for changes at `path` calling `callback`
+ * @param {string}
+ * @param {function|object=} [options]
+ * @param {string=} [options.encoding = 'utf8']
+ * @param {AbortSignal=} [options.signal]
+ * @return {Watcher}
+ */
+export function watch (path, options) {
```
chicoxyzzy, line 497: params do not match in jsdoc and function notation

api/fs/dir.js
```diff
@@ -60,12 +60,30 @@ export class Dir {
...
+    return Boolean(this.handle?.closed)
+  }
+
+  /**
+   * `true` if closeing, otherwise `false`.
+   * @ignore
+   * @type {boolean}
+   */
+  get closing () {
+    return Boolean(this.handle?.closig)
```
heapwolf, line 78: typo

api/fs/watcher.js
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+ * Starts the `fs.Watcher`
+ * @ignore
+ * @param {Watcher} watcher
+ * @return {Promise}
+ */
+async function start (watcher) {
+  const result = await ipc.send('fs.watch', {
+    path: watcher.path,
```
heapwolf, line 36: should we double check these parameters exist? defaulting to something here seems like a possible foot-gun.

jwerle: they come from the `Watcher` instance - I can verify there because this function is private

heapwolf (approved):

chicoxyzzy (approved):

api/fs/watcher.js
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * A container for a file system path watcher.
+ */
+export class Watcher extends EventEmitter {
+  /**
+   * The underlying `fs.Watcher` resource id.
+   * @ignore
+   * @type {string}
+   */
+  id = null
```
bcomnes, line 85: Random TS question:

If you don't set this to null in the instance value field, and you guarantee set them in the constructor, typescript will let you type it as never null and just the value its typed as. Or maybe this is a strict null check option, I'm not sure. Typescript is a million variants of itself depending on how you configure it. 

What about not setting this to null initially, or if there is a reason, why do we set this to null here?

jwerle: just by habit
@chicoxyzzy what do you think we should set it to here?

bcomnes (approved):
Nice

api/fs/watcher.js
bcomnes: Not setting null here (unless I'm mistaken), and also setting in the constructor narrows the fields slightly to be a single type instead of also possibly null.

PR #746, opened 2023-11-03
Implement WebView Navigator File System Mounts for desktop

### LINKED ISSUE #715: File System Mounts
As we move to sandboxing for all applications, we'll need a way to provide file system access with permission from the end-user which Apple makes this possible with [`startAccessingSecurityScopedResource`](https://developer.apple.com/documentation/foundation/nsurl/1417051-startaccessingsecurityscopedreso?language=objc) and [`bookmarkDataWithContentsOfURL:error:`](https://developer.apple.com/documentation/foundation/nsurl/1408344-bookmarkdatawithcontentsofurl?language=objc) and Widows make it possible with [Windows sandbox configuration](https://learn.microsoft.com/en-us/windows/security/application-security/application-isolation/windows-sandbox/windows-sandbox-configure-using-wsb-file)

## File System Mounts

A mechanism for configuring a mounted file system path on the host that is available to the through the navigator URI at socket://<bundle_identifier>/mnt/<mount>

### `socket.ini` Configuration

```ini
[navigator.mounts]
$HOST_HOME/.local/share/my-app = "user-shared"
$HOST_CONTAINER/shared = "app-shared"
```

The `[navigator.mounts]` section will have access to special variables:

```ini
# The HOME directory of the user host
HOST_HOME = "/Users/jwerle"
# The container directory for the application with respect to the host user 
HOST_CONTAINER = "/Users/jwerle/Library/Containers/com.company.app"
# The working directory of the application process
HOST_PROCESS_WORKING_DIRECTORY = "/Users/jwerle/Applications/My App.app/Contents/MacOS/Resources"
```

The application will mount the paths and make them available in the navigator:

```js
const plugin = await import('/mnt/user-shared/plugin.js')
```

We could provide runtime mounts with an API:

```js
import { mount } from 'socket:application'
await mount('$HOST_HOME/.local/shared/my-app, 'user-shared')
const plugin = await import('/mnt/user-shared/plugin.js')
```

jwerle:
Closes #715

### DIFF
--- a/api/CONFIG.md
+++ b/api/CONFIG.md
@@ -62,6 +62,14 @@ root | "/" |  Make root open index.html
 default_index | "" |  Set default 'index.html' path to open for implicit routes
 watch | false |  Enable watch mode
 
+# `webview.navigator.mounts`
+
+Key | Default Value | Description
+:--- | :--- | :---
+$HOST_HOME/directory-in-home-folder/ |  | 
+$HOST_CONTAINER/directory-app-container/ |  | 
+$HOST_PROCESS_WORKING_DIRECTORY/directory-in-app-process-working-directory/ |  | 
+
 # `permissions`
 
 Key | Default Value | Description
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -4583,8 +4583,13 @@ int main (const int argc, const char* argv[]) {
         );
       }
 
-      if (settings["ios_sandbox"] != "false") {
+      if (flagDebugMode) {
+        entitlementSettings["configured_entitlements"] += (
+          "  <key>get-task-allow</key>\n"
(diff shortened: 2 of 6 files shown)


api/CONFIG.md
```diff
@@ -62,6 +62,14 @@ root | "/" |  Make root open index.html
 default_index | "" |  Set default 'index.html' path to open for implicit routes
 watch | false |  Enable watch mode
 
+# `webview.navigator.mounts`
```
jwerle, line 65: @chicoxyzzy this looks weird - maybe I should just omit the sample values...

chicoxyzzy: what's weird? it looks good to me

bcomnes (approved):

Issue #716, opened 2023-10-24
Replace arguments passed to the [build] script by environment variable


chicoxyzzy:
`[build] script` adds arguments that we pass to the script. This could break other tools so this requires some really hacky tricks ([like this](https://github.com/socketsupply/create-socket-app/pull/55/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R104-R107)). We should move to envs instead

chicoxyzzy:
this one is now replaced by #743

PR #747, opened 2023-11-03
Introduce '[build] build.forward_arguments'

chicoxyzzy:
This is the simplest solution which fixes the #716. It doesn't pass any arguments as environment variables and just skips them until `[build] pass.build.arguments` is equal `true` explicitely

### DIFF
--- a/api/CONFIG.md
+++ b/api/CONFIG.md
@@ -47,7 +47,7 @@ headless | false |  If true, the window will never be displayed.
 name |  |  The name of the program and executable to be output. Can't contain spaces or special characters. Required field.
 output | "build" |  The binary output path. It's recommended to add this path to .gitignore.
 script |  |  The build script. It runs before the `[build] copy` phase.
-pass.build.arguments | false |  If true, it will pass build arguments to the build script. WARNING: this could be deprecated in the future.
+build.script.forward_arguments | false |  If true, it will pass build arguments to the build script. WARNING: this could be deprecated in the future.
 
 # `build.watch`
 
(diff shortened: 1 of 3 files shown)


jwerle (approved):

api/CONFIG.md
```diff
@@ -47,6 +47,7 @@ headless | false |  If true, the window will never be displayed.
 name |  |  The name of the program and executable to be output. Can't contain spaces or special characters. Required field.
 output | "build" |  The binary output path. It's recommended to add this path to .gitignore.
 script |  |  The build script. It runs before the `[build] copy` phase.
+pass.build.arguments | false |  If true, it will pass build arguments to the build script. WARNING: this could be deprecated in the future.
```
chicoxyzzy, line 50: This name is discussable. I'm open to another opinions.

@jwerle I made the separator a `.` here. Is it ok or should I revert to `_` for the `0.5` release?

jwerle: The `.` is correct. We should probably call this property `build.script.forward_arguments`

src/cli/cli.cc
```diff
@@ -507,18 +506,16 @@ void handleBuildPhaseForUserScript (
 #endif
   } while (0);
 
-  StringStream buildArgs;
-  buildArgs << " " << pathResourcesRelativeToUserBuild.string();
-  buildArgs << " " << additionalArgs;
+  const bool shouldPassBuildArgs = settings.contains("build_build.script.forward_arguments") && settings.at("build_build.script.forward_arguments") == "true";
```
jwerle, line 509: this is still wrong. It needs to be `build.script.forward_arguments`

jwerle (approved):

PR #764, opened 2023-11-10
MIME-DB tests

bcomnes:
~~WIP. These asserts still need work probably.~~ This is ready for review!


jwerle (approved):

test/src/mime.js
```diff
@@ -1,8 +1,132 @@
...
+  }
+
+  for (const item of right) {
+    if (left.includes(item)) {
+      // @ts-ignore
+      results.add(item)
+    }
+  }
+
+  return Array.from(results.entries())
```
chicoxyzzy, line 25: is it correct that it's an array of entries and not values?

jwerle: `Set.prototype.entries()` returns an [`Iterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator) (_Set Entries_) that is an [`Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols), which we convert to an array with `Array.from()`. This is desired

chicoxyzzy: The return value is in JSDoc is wrong than, it's `[number, string][]` on this line and `string[]` in JSDoc

chicoxyzzy: maybe we can simplify it to
```js
return left.filter(value => right.includes(value));
```
?

chicoxyzzy (changes requested):
the `intersection` implementation looks incorrect

chicoxyzzy (approved):

PR #771, opened 2023-11-15
update ssc init

chicoxyzzy:
make `<path>` work, take name from path when possible

@jwerle probably same changes needed in the CLI refactor branch.

- Project name will be the last child directory name (current directory name if `ssc init` is run without a path)
- `--name` have preference over the one from the path


chicoxyzzy:
this is an important bugfix

jwerle (approved):

PR #777, opened 2023-11-20
Smart 'dgram.Socket' reuse

jwerle:
This PR introduces a smart mechanism for reusing underlying socket handles by deriving the underlying `id` from the `host` and `port` when `port !== 0` and `reuseAddr === true` such that an already known `id` was not given to `createSocket()`. The implementation uses the `murmur3` non cryptographic hashing algorithm based on https://github.com/jwerle/murmurhash.c. It uses the current window index as the `seed` value and the concatenated `host + port` tuple as an input value. This PR depends on #776


test/scripts/test-desktop.js
```diff
@@ -25,7 +25,7 @@ cp(path.join(root, 'fixtures'), path.join(TMPDIR, 'ssc-socket-test-fixtures'), {
 })
 
 try {
-  exec(`ssc build -r --test=./index.js ${!DEBUG ? '-o --headless --prod' : ''}`, {
+  exec(`ssc build -r --test=./index.js ${!DEBUG ? '-o  --prod' : ''}`, {
```
chicoxyzzy, line 28: why drop `--headless`?

jwerle: its already in `socket.ini` - if we want to debug without headless on desktop, we need to comment it out in the `socket.ini` file and then remove it here

PR #789, opened 2023-11-21
Add dragdropfiles event

bcomnes:
Dispatch dragdropfiles with FileSystemDirectoryHandle or FileSystemFileHandle entires in a handles array on the event.

### DIFF
--- a/api/internal/init.js
+++ b/api/internal/init.js
@@ -79,12 +79,16 @@ if ((globalThis.window) === globalThis) {
   globalThis.addEventListener('dragdropfiles', async (event) => {
     const { files } = event.detail
     const handles = []
-    for (const file of files) {
-      const stats = await fs.stat(file)
-      if (stats.isDirectory()) {
-        handles.push(await createFileSystemDirectoryHandle(file, { writable: false }))
-      } else {
-        handles.push(await createFileSystemFileHandle(file, { writable: false }))
+    if (Array.isArray(files)) {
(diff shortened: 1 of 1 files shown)


api/internal/init.js
```diff
@@ -69,6 +74,24 @@ if ((globalThis.window || globalThis.self) === globalThis) {
   }
 }
 
+// webview only features
+if ((globalThis.window) === globalThis) {
+  globalThis.addEventListener('dragdropfiles', async (event) => {
+    const { files } = event.detail
+    const handles = []
+    for (const file of files) {
```
jwerle, line 82: Can we add an Array.isArray check 🙏🏼

api/internal/init.js
```diff
@@ -69,6 +74,24 @@ if ((globalThis.window || globalThis.self) === globalThis) {
...
+      const stats = await fs.stat(file)
```
jwerle, line 83: Can we add a `typeof file === string` check?

api/internal/init.js
bcomnes: Done

api/internal/init.js
bcomnes: Done

bcomnes:
@jwerle I think that covers your requests?

jwerle (approved):

chicoxyzzy (approved):

PR #802, opened 2023-11-23
fix(src/cli/cli.cc): fix headless mode

chicoxyzzy:


chicoxyzzy:
@jwerle this change could be necessary in a new CLI branch

jwerle (approved):

PR #811, opened 2023-11-28
add ios nonexempt encryption default

heapwolf:
includes a minor copy update to the ini file

### DIFF
--- a/src/cli/templates.hh
+++ b/src/cli/templates.hh
@@ -1810,9 +1810,11 @@ distribution_method = "ad-hoc"
 ; A path to the provisioning profile used for signing iOS app.
 provisioning_profile = ""
 
-; which device to target when building for the simulator
+; which device to target when building for the simulator.
 simulator_device = "iPhone 14"
 
+; Indicate to Apple if you are using encryption that is not exempt.
+nonexempt_encryption = "false"
 
 [linux]
 ; Helps to make your app searchable in Linux desktop environments.
--- a/src/cli/templates.hh
+++ b/src/cli/templates.hh
@@ -1165,6 +1165,8 @@ constexpr auto gIOSInfoPList = R"XML(<?xml version="1.0" encoding="UTF-8"?>
   <key>NSSupportsAutomaticGraphicsSwitching</key>
   <true/>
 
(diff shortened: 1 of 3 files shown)


src/cli/templates.hh
```diff
@@ -1165,6 +1165,8 @@ constexpr auto gIOSInfoPList = R"XML(<?xml version="1.0" encoding="UTF-8"?>
   <key>NSSupportsAutomaticGraphicsSwitching</key>
   <true/>
 
+  <key>ITSAppUsesNonExemptEncryption</key>
+  <{{ios_nonexempt_encryption}}/>
```
chicoxyzzy, line 1169: Do you also want to add it to the config template to appear in the docs?

src/cli/templates.hh
```diff
@@ -1613,14 +1615,15 @@ constexpr auto gDefaultConfig = R"INI(
...
+; a shell command when  you run "ssc build". This is the most important command
+; in this file. It will do all the heavy lifting and should handle 99.9% of your
+; use cases for moving files into place or tweaking platform-specific build
+; artifacts. If you don't specify it, ssc will just copy everything in your
+; project to the build target.
 ;
-; Note that "~" alias won't expand to the home directory in any of the configuration files.
-; Use the full path instead.
+; Note that "~" alias won't expand to the home directory in any of the config
+; files. Use the full path instead.
```
chicoxyzzy, line 1626: Hmm the CI should fail here because the CONFIG.md was not regenerated with `npm run gen`, but it passes 🤔

src/cli/templates.hh
```diff
@@ -1807,9 +1810,11 @@ distribution_method = "ad-hoc"
 ; A path to the provisioning profile used for signing iOS app.
 provisioning_profile = ""
 
-; which device to target when building for the simulator
+; which device to target when building for the simulator.
 simulator_device = "iPhone 14"
 
+; Indicate to Apple if you are using encryption that is not exempt.
+nonexempt_encryption = "false"
```
chicoxyzzy, line 1817:
```suggestion
; Indicate to Apple if you are using encryption that is not exempt.
; default value: false
; nonexempt_encryption = false
```

chicoxyzzy (approved):

PR #814, opened 2023-11-28
fix(src/ipc/bridge.cc): escape mount mapping keys

### LINKED ISSUE #810: Empty imports on filesystem mounted modules
With the following config in `socket.ini`:

```ini
[webview.navigator.mounts]
$HOST_HOME/.local/relay/plugins = "/relay-plugins"
```

Where the tree of `~/.local/relay/plugins` looks like:

```bash
.
├── plugin-1
│   └── plugin.js
└── plugin-2
    └── plugin.js
```

Example `plugin.js`

```js
export default async function plugin1({
  appView
}) {
  console.log('plugin1 Loaded')
  console.log({
    name: plugin1,
    appView
  })
}
```

When I import the following URLs:

```js
const plugin = await import('/relay-plugins/plugin-2/plugin.js')
```

Plugin comes back as an empty object, when it should have a default function in it. 

```
plugin: {}
```

jwerle:
Fixes #810

### DIFF
--- a/src/ipc/bridge.cc
+++ b/src/ipc/bridge.cc
@@ -3750,10 +3750,10 @@ namespace SSC::IPC {
     }
 
     Map mappings = {
-      {"$HOST_HOME", HOME},
+      {"\\$HOST_HOME", HOME},
       {"~", HOME},
 
-      {"$HOST_CONTAINER",
+      {"\\$HOST_CONTAINER",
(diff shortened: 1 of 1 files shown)


bcomnes (approved):

chicoxyzzy (approved):

bcomnes:
Nice, will try this out now.

PR #821, opened 2023-12-04
Fix `FileSystemWatcher` high CPU

jwerle:
This PR addresses the internal implementation of the `FileSystemWatcher` class by replacing a custom event loop implementation with one provided by `SSC::Core`


heapwolf (approved):
looks good, minor nit

PR #837, opened 2023-12-07
Implement JS handler for ApplicationURLEvent

bcomnes:
Note this is a PR against the `[application-url-hook](https://github.com/socketsupply/socket/tree/application-url-hook)` branch.

### DIFF
--- a/api/hooks.js
+++ b/api/hooks.js
@@ -57,6 +57,10 @@
  * hooks.onNotificationPresented((event) => {
  *   // called when 'notificationpresented' events are dispatched on the global object
  * })
+ *
+ * hooks.onNotificationPresented((event) => {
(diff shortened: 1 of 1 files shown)


api/hooks.js
```diff
@@ -111,6 +115,8 @@ function proxyGlobalEvents (global, target) {
         dispatchEvent(target, new MessageEvent(type, { origin, data, detail }))
       } else if (detail) {
         dispatchEvent(target, new CustomEvent(type, { detail }))
+      } else if (detail) {
```
chicoxyzzy, line 118: this branch will never execute

bcomnes: It was supposed to be URL, but I think I Ctl-z'd it back to detail.

PR #841, opened 2023-12-11
ApplicationURLEvent

### LINKED ISSUE #568: Fix deb package name and meta for `linux-arm64`
it's always `amd64`, should be `arm64` (?) for `linux-arm64`
https://github.com/socketsupply/socket/tree/95b8e872248f6057699915fca2e203fc2c69cf5d/src/cli/cli.cc#L1587 
https://github.com/socketsupply/socket/tree/95b8e872248f6057699915fca2e203fc2c69cf5d/src/cli/templates.hh#L341

### LINKED ISSUE #790: Configuration and API for Custom Application Protocol URL Scheme
In an effort to allow applications built with the runtime to respond to "deep links", a configuration and API for custom application protocol URL schemes should exist.

## Configuration

The `socket.ini` configuration COULD look like this:

```ini
[meta.application]
protocol = "custom-scheme"
```

## JavaScript API

The API for handling "deep links" COULD be exposed in the `socket:hooks` module with an `ApplicationURLEvent`

```js
import hooks from 'socket:hooks'

hooks.onApplicationURL((/* @type {ApplicationURLEvent} */ event) => {
  const { url, data } = event

  console.log(url.toString()) // custom-scheme://action?key=value
  console.log(data) // nullable arbitrary data
})
```

The `ApplicationURLEvent` could be a non-standard `Event` class extension:

```js
new ApplicationURLEvent(type = 'applicationurl') // possible, but useless
new ApplicationURLEvent(type, { url: string|URL, data?: object })
```

With an implementation like:

```js
class ApplicationURLEvent extends Event {
  #data = null
  #url = null

  constructor (type, options = null) {
    super(type)
    this.#data = options?.data ?? null
    this.#url = options?.url ?? null
    this[Symbol.toStringTag] = 'ApplicationURLEvent'
  }

  get isTrusted () { return true }
  get data () { return this.#data ?? null }
  get url () { return this.#url ? new URL(this.#url) : null }
}
```

jwerle:
Closes #790 
Closes #568


src/android/main.kt
```diff
@@ -174,8 +184,15 @@ open class MainActivity : WebViewActivity() {
     super.onNewIntent(intent)
     val window = this.window
     val action = intent.action
+    val data = intent.data
     val id = intent.extras?.getCharSequence("id")?.toString()
 
+    if (data != null) {
+      window.bridge.emit("applicationurl", """{
+        "url": "$data"
```
heapwolf, line 192: is this 100% safe?

heapwolf (approved):

src/android/main.kt
jwerle: in fact, no it isn't! I need to add a check that the action is either `"android.intent.action.MAIN"` or `"android.intent.action.VIEW"`. We can include a check that the `data` is prefixed with the register application protocol scheme configured in the `socket.ini`

src/desktop/main.cc
```diff
@@ -116,6 +184,137 @@ MAIN {
...
+      Thread thread([connection] () {
+        while (dbus_connection_read_write_dispatch(connection, 100));
+        app_ptr->dispatch(pollForMessage);
+      });
+
+      thread.detach();
+    };
+
+    app_ptr->dispatch(pollForMessage);
+  } else if (argc > 1 && String(argv[1]).starts_with(appProtocol + ":")) {
```
chicoxyzzy, line 243: should we check that `appProtocol` is a non-empty string for sanity?

src/cli/templates.hh
```diff
@@ -269,6 +269,18 @@ constexpr auto gMacOSInfoPList = R"XML(<?xml version="1.0" encoding="UTF-8"?>
...
+  <array>
+    <dict>
+      <key>CFBundleURLName</key>
+      <string>{{meta_application_protocol}}</string>
+      <key>CFBundleURLSchemes</key>
+      <array>
+        <string>{{meta_application_protocol}}</string>
+      </array>
+    </dict>
+  </array>
```
chicoxyzzy, line 282: will this work correctly if `meta_application_protocol`is an empty string?

src/desktop/main.cc
```diff
@@ -116,6 +184,137 @@ MAIN {
   bool wantsVersion = false;
   bool wantsHelp = false;
 
+  auto bundleIdentifier = app.appData["meta_bundle_identifier"];
```
chicoxyzzy, line 187: it's already defined on line 503

chicoxyzzy (changes requested):
macOS build fails, see my comments above

src/desktop/main.cc
chicoxyzzy: Can we make the value getters consistent, i.e. `userConfig["meta_bundle_identifier"]` vs `app.appData["meta_bundle_identifier"]`?

bcomnes (approved):

src/cli/cli.cc
```diff
@@ -2022,7 +2022,7 @@ int main (const int argc, const char* argv[]) {
         settings["build_name"] + "_" +
         "v" + settings["meta_version"] + "-" +
         settings["meta_revision"] + "_" +
-        replace(platform.arch, "x86_64", "amd64")
+        replace(SSC::platform.arch, "x86_64", "amd64")
```
chicoxyzzy, line 2025: just linking https://github.com/socketsupply/socket/issues/568 here, so we don't forget to fix it here later as well

src/cli/cli.cc
```diff
@@ -2592,7 +2592,7 @@ int main (const int argc, const char* argv[]) {
       Path pathPackage = paths.platformSpecificOutputPath / (
         settings["build_name"] + "_" +
         settings["meta_version"] + "_" +
-        replace(platform.arch, "x86_64", "amd64")
+        replace(SSC::platform.arch, "x86_64", "amd64")
```
chicoxyzzy, line 2595: just linking https://github.com/socketsupply/socket/issues/568 here, so we don't forget to fix it here later as well

chicoxyzzy (approved):

PR #846, opened 2023-12-13
fix(src/android/runtime.cc): fix Android runtime

chicoxyzzy:


jwerle (approved):

chicoxyzzy:
@jwerle pls check the last patch

PR #848, opened 2023-12-13
fs tests

bcomnes:
Pushing these up as I get them working.

### DIFF
--- a/test/src/fs/index.js
+++ b/test/src/fs/index.js
@@ -5,6 +5,7 @@ import crypto from 'socket:crypto'
 import path from 'socket:path'
 import fs from 'socket:fs'
 import os from 'socket:os'
+import process from 'socket:process'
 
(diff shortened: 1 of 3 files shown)


test/src/fs/index.js
```diff
@@ -47,7 +48,7 @@ test('fs.access', async (t) => {
   })
 })
 
-test('fs.appendFile', async (t) => {})
+// fs.appendFile not implemented. wontFix
```
jwerle, line 51: we can just remove this line

jwerle (approved):

bcomnes:
Ok, I have a few more tests to write, but this file generally needs a bunch of cleanup. Also many of the different parts of the API remain unexercised. Any thoughts on coverage tooling yet?

bcomnes:
Lets just land this as-is and add any further improvements in another branch.

heapwolf (approved):

PR #849, opened 2023-12-14
feat: Experimental WASM Extensions

jwerle:
This PR introduces support for WASM C/C++ extensions that support a subset of libc and the Socket Runtime Native Extension API.

Targeting `wasm32` extensions is experimental as the ABI and APIs may be incomplete or subject to change

WASM extensions can be configured like:

```ini
[meta]
name = "my-wasm-extension"
type = "extension"

[extension]
target = "wasm32"
sources[] = extension.cc
```

Implemented like:

```c++
#include <socket/extension.h>
#include <string.h>

static void onEcho (
  sapi_context_t* context,
  sapi_ipc_message_t* message,
  const sapi_ipc_router_t* router
) {
  auto result = sapi_ipc_result_create(context, message);
  auto value = sapi_ipc_mesage_get(message, "value");
  auto data = sapi_json_string_create(context, value);
  sapi_ipc_result_set_json_data(result, sapi_json_any(data));
  sapi_ipc_reply(result);
}

static bool initialize (sapi_context_t* context, const void* data) {
  sapi_ipc_router_map(context, "my-wasm-extension.echo", onEcho, NULL);
  return true;
}

SOCKET_RUNTIME_REGISTER_EXTENSION("my-wasm-extension", initialize);
```

Consumed like this:

```js
import extension from 'socket:extension'

const wasm = await extension.load('my-wasm-extension')
const result = await wasm.binding.echo({ value: 'hello' })
console.log(result.data) // 'hello'
```


chicoxyzzy (approved):

chicoxyzzy (changes requested):
macOS CI is failing with
```
error: unable to create target: 'No available targets are compatible with triple "wasm32"'
```

chicoxyzzy (approved):

PR #853, opened 2023-12-18
Enable bluetooth on demand guarded by permissions

### LINKED ISSUE #644: permissions.allow_bluetooth does nothing
### What version Socket Runtime are you using?

v0.4.0

### What did you expect to see and what you saw instead?

It appears that `allow_bluetooth = false` in the `[permissions]` section is a no-op. In fact, every permission is a no-op. Will I be able to use this to disable the Bluetooth feature of Socket in the future?

jwerle:
Fixes #644 

cc @aleclarson

### DIFF
--- a/src/core/bluetooth.cc
+++ b/src/core/bluetooth.cc
@@ -430,10 +430,6 @@ using namespace SSC;
 
 namespace SSC {
   Bluetooth::Bluetooth () {
-    #if defined(__APPLE__)
-    this->controller = [SSCBluetoothController new];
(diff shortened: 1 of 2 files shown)


chicoxyzzy:
I've tested with

```ini
[permissions]
allow_bluetooth = false
```

I don't see the Bluetooth pop-up window anymore, but I also can't see it with `allow_bluetooth = true`.
I also can do `await import('socket:bluetooth')` when `allow_bluetooth = false`

PR #856, opened 2023-12-19
fix(api/path/path.js): avoid static initializer

chicoxyzzy:
Static initialization blocks are only supported since Safari 16.4, so users on macOS Ventura will get an error

image


jwerle (approved):

chicoxyzzy:
this fixes the issue for @lamiazar

PR #868, opened 2023-12-28
Transparent window background

chicoxyzzy:
- [x] macOS
- [ ] Linux
- [ ] Windows

### DIFF
--- a/src/android/window.cc
+++ b/src/android/window.cc
@@ -44,6 +44,7 @@ namespace SSC::android {
     const auto argv = this->config["ssc_argv"];
 
     options.headless = this->config["build_headless"] == "true";
+    options.transparent = this->config["build_transparent"] == "true";
     options.debug = isDebugEnabled() ? true : false;
     options.env = stream.str();
     options.cwd = rootDirectory.str();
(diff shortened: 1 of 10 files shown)


src/android/window.cc
```diff
@@ -44,6 +44,7 @@ namespace SSC::android {
     const auto argv = this->config["ssc_argv"];
 
     options.headless = this->config["build_headless"] == "true";
+    options.transparent = this->config["build_transparent"] == "true";
```
chicoxyzzy, line 47: oh I guess that this is useless for Android

chicoxyzzy:
I'm trying to add this on Windows
```c++
    if (opts.transparent) {
      CORE2WEBVIEW_COLOR color = { 0 };
      webview->put_DefaultBackgroundColor(color);
    }
```
it compiles fine, but in result I see that app height and weight are 0px, like something is going wrong. When I remove or comment lines above, it compiles fine. It's weird that there're no errors in the app, but it's just broken

I took example code from here
https://github.com/MicrosoftEdge/WebView2Samples/blob/85e53c9dc4245ed74fc737554640644d4b0f3bb5/SampleApps/WebView2APISample/ViewComponent.cpp#L84
and it looks correct acccording to what I see in the webview2 win32 c++ api

chicoxyzzy:
TODO: add tests for JS API

chicoxyzzy:
Should I rebase the branch? cc @heapwolf @jwerle

PR #882, opened 2024-01-05
Fix 'util.isArgumentsObject' predicate

jwerle:


api/util.js
```diff
@@ -48,7 +48,11 @@ export function isAsyncFunction (object) {
 }
 
 export function isArgumentsObject (object) {
-  return isPlainObject(object)
+  return (
+    !Array.isArray(object) &&
+    isPlainObject(object) &&
+    Number.isFinite(object.length)
```
chicoxyzzy, line 54: Just curious, why is las check necessary? Also it could be more strict probably (`Number.isInteger`)

chicoxyzzy (approved):

PR #883, opened 2024-01-05
Improve global env

jwerle:


api/internal/monkeypatch.js
```diff
@@ -169,7 +170,13 @@ export function init () {
...
     FileSystemFileHandle,
     FileSystemDirectoryHandle,
-    FileSystemWritableFileStream
+    FileSystemWritableFileStream,
+
+    // buffer
+    Buffer,
+
+    // platform detection
+    isSocketRuntime: true
```
chicoxyzzy, line 179: we can also make it non-configurable, but it's not very important I guess

test/src/webview.js
```diff
@@ -1,11 +1,111 @@
...
+test('globalThis.onload', async (t) => {
+  t.ok(await loaded, 'globalThis.onload called')
+})
+
+test('globalThis.isSocketRuntime', (t) => {
+  t.ok(globalThis.isSocketRuntime === true, 'globalThis.isSocketRuntime')
+})
+
+test('globalThis.Buffer', (t) => {
+  t.ok(typeof globalThis.Buffer === 'function' && globalThis.Buffer === Buffer, 'globalThis.Buffer')
```
chicoxyzzy, line 38: I'd change the message here and in other places to something like "Buffer is available on globalThis" and "globalThis.URL is normalized"

chicoxyzzy (approved):

test/src/webview.js
jwerle: we can improve this later 🙏🏼

api/internal/monkeypatch.js
jwerle: The `install()` function already does that

PR #892, opened 2024-01-11
feat(socket:ip): Introduce 'socket:ip' module

heapwolf:
This PR introduces the `socket:ip` module and removes the useless `socket:net` module


chicoxyzzy (approved):
this needs `npm run gen` to regenerate types

api/ip.js
```diff
@@ -0,0 +1,13 @@
...
+ * @module IP
+ * Functions for working with IP addresses
+ *
+ */
+
+const v4Seg = '(?:[0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])'
+const v4Str = `(${v4Seg}[.]){3}${v4Seg}`
+const IPv4Reg = new RegExp(`^${v4Str}$`)
+
+export const isIPv4 = s => {
```
jwerle, line 11: can we rewrite this as:

```js
/**
 * Determines if an input `string` is in IP address version 4 format.
 * @param {string} string
 * @return {boolean}
 * /
export function isIPv4 (string) {
  return typeof string === 'string' && IPv4Reg.test(string)
}
```

api/ip.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @module IP
+ * Functions for working with IP addresses
+ *
```
jwerle, line 4: new line

jwerle (changes requested):
Left some comments. We need to generate docs for this change: `npm run gen`

api/ip.js
```diff
@@ -0,0 +1,13 @@
+/**
```
jwerle, line 1: This module needs a `default` exports as all `socket:` modules require it:

```js
export default { isIPv4 }
```

should work

jwerle (approved):

PR #912, opened 2024-02-03
Support Llama.cpp as a core module

heapwolf:
This PR will add JavaScript bindings for llama.cpp

```js
import { Model, Context, Grammar } from 'socket:llm'

// include a model file in your project
const model = new Model('./mistral-7b-openorca.Q4_0.gguf')

// create a grammar
const grammar = new Grammar('...') // supports (json)gbnf

// create a context
const context = new Context({ model, grammar })

// use the context to evaluate the input and produce output
context.eval(tokens, params)
```


ansarizafar:
Google has recently released gemma models and https://github.com/google/gemma.cpp.

heapwolf:
added to #next branch

PR #915, opened 2024-02-04
Custom webview headers and applinks

### LINKED ISSUE #871: Support configurable window navigation response headers
This should allow for [navigation headers](https://developer.mozilla.org/en-US/docs/Glossary/Response_header) to be set, such as the [`frame-ancestors` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors)

We need to add response navigation headers in the request response for HTML loaded in the WebView:

## iOS & macOS
- Add headers to the loaded [`NSMutableURLRequest`](https://developer.apple.com/documentation/foundation/nsmutableurlrequest?language=objc)

## Windows
- Add headers in the [`WebResourceRequested`](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2webresourcerequest?view=webview2-1.0.2210.55) event

## Linux
- Construct [`WebKitURIRequest`](https://webkitgtk.org/reference/webkit2gtk/stable/class.URIRequest.html)
- Migrate navigation (`Window::navigate()`) to use  [`webkit_web_view_load_request`](https://webkitgtk.org/reference/webkit2gtk/stable/method.WebView.load_request.html) instead of [`webkit_web_view_load_uri`](https://webkitgtk.org/reference/webkit2gtk/stable/method.WebView.load_uri.html)
- Add headers with [`soup_message_headers_append `](https://libsoup.org/libsoup-3.0/method.MessageHeaders.append.html) to the `SoupMessageHeaders*` headers returned from [`webkit_uri_request_get_http_headers`](https://webkitgtk.org/reference/webkit2gtk/stable/method.URIRequest.get_http_headers.html)

## Android
- Rework how initial HTML is loaded by using [`loadUrl`](https://developer.android.com/reference/kotlin/android/webkit/WebView#loadurl) instead of [`loadDataWithBaseURL `](https://developer.android.com/reference/kotlin/android/webkit/WebView#loaddatawithbaseurl)
- In order to ensure that the preload is evaluated in the top most part of the document, the `shouldInterceptRequest` callback method should read the entire loaded HTML (if requested) and inject it, and then return the `WebResourceResponse`

## `socket.ini`

```ini
[webview]
headers[] = "Content-Security-Policy: frame-ancestors <source>;"
headers[] = "X-Custom-Header: Some-Value"
```

We'll likely need to handle the `[webview] headers` value in the configuration as an exception so each value in the array will end with a `\n` if a value does not already end with one so these can be set correctly and the parser doesn't confuse `[webview] headers` as a "space delimited" list of values

jwerle:
Closes #871 

This enables setting custom HTTP headers for all navigator requests where

```ini
[webview]
headers[] = "X-Custom-Header: some-value"
```

will be made available in all requests made in the webview.

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -6870,7 +6870,8 @@ int main (const int argc, const char* argv[]) {
           sourcesWatcher->options.debounce = std::atoi(timeout.c_str());
         } catch (Exception e) {
           log(
-            "ERROR: Invalid value given for '[build.watch] debounce_timeout': " + timeout
+            "ERROR: Invalid value given for '[build.watch] debounce_timeout': '" + timeout + "'. " +
+            "Expecting an integer value in milliseconds"
           );
           exit(1);
         }
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -509,9 +509,9 @@ void handleBuildPhaseForUserScript (
   do {
     char prefix[4096] = {0};
     std::memcpy(
-    prefix,
-    pathResourcesRelativeToUserBuild.string().c_str(),
-    pathResourcesRelativeToUserBuild.string().size()
+      prefix,
+      pathResourcesRelativeToUserBuild.string().c_str(),
+      pathResourcesRelativeToUserBuild.string().size()
     );
 
     // @TODO(jwerle): use `Env::set()` if #148 is closed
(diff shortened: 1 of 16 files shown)


src/cli/cli.cc
```diff
@@ -6776,12 +6863,28 @@ int main (const int argc, const char* argv[]) {
...
 
       sourcesWatcher = new FileSystemWatcher(sources);
+
+      if (settings["build_watch_debounce_timeout"].size() > 0) {
+        const auto timeout = settings["build_watch_debounce_timeout"];
+        try {
+          sourcesWatcher->options.debounce = std::atoi(timeout.c_str());
+        } catch (Exception e) {
+          log(
+            "ERROR: Invalid value given for '[build.watch] debounce_timeout': " + timeout
```
heapwolf, line 6873: should we specify an expectation that includes the unit of measurement?

src/cli/templates.hh
```diff
@@ -267,7 +267,7 @@ constexpr auto gMacOSInfoPList = R"XML(<?xml version="1.0" encoding="UTF-8"?>
   <string>MainMenu</string>
 
   <key>NSPrincipalClass</key>
-  <string>AtomApplication</string>
```
heapwolf, line 270: 🧐

heapwolf (approved):

PR #916, opened 2024-02-04
Fix and improve drag and drop API

jwerle:
This PR fixes drag and drop for DOM elements and improves dropping files into the webview by implementing and supporting the DataTransfer API in the `drop` event. Consumers of the `drop` event can use `event.dataTransfer.items` or `event.dataTransfer.files`. File and directory handles can be accessed with the new [`getAsFileSystemHandle`](https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/getAsFileSystemHandle) API on `DataTransferItem` instances available in `drop` events. This method is otherwise undefined in other contexts where a `DataTransferItem` instance exists.

an example of handling dropped in files and directories:

```js
target.addEventListener('drop', async (event) => {
  for (const item of event.dataTransfer.items) {
    if (item.kind === 'file' || item.type === 'text/directory') {
      const handle = await item.getAsFileSystemHandle()
      if (handle.kind === 'file') {
        const arrayBuffer = await handle.getFile().arrayBuffer()
      } else if (handle.kind === 'directory') {
        for await (const [filename] of handle.entries() {
          console.log(filename)
        }
      }
    }
  }
})
```


api/internal/init.js
```diff
@@ -76,11 +76,20 @@ if ((globalThis.window || globalThis.self) === globalThis) {
 
 // webview only features
 if ((globalThis.window) === globalThis) {
-  globalThis.addEventListener('dragdropfiles', async (event) => {
-    const { files } = event.detail
+  globalThis.addEventListener('platformdrop', async (event) => {
```
chrisfarms, line 79: probably fine in this context, but this `async` func has no catch, and being passed to something that doesn't handle the promise right? ... so I think it is gonna swallow errors or trigger a confusing unhandled-promise error if something goes wrong

heapwolf (approved):
this is an awesome improvement. 🎉

PR #924, opened 2024-02-08
s/ssrc/sscrc

bcomnes:
I think `ssrc` might be a typo. When I use `.ssrc` the override file doesn't work. When I use `.sscrc` it works.


jwerle (approved):

bcomnes:
Fixed, sorry for the oversight.

PR #925, opened 2024-02-08
refactor(api/window/hotkey.js): dedupe global hot key events

jwerle:


api/window/hotkey.js
```diff
@@ -162,6 +167,12 @@ export class Bindings extends EventTarget {
     const { id } = event.data ?? {}
     const binding = this.get(id)
 
+    if (event.timeStamp < this.#clock) {
```
heapwolf, line 170: is timeStamp always defined?

heapwolf (approved):

api/window/hotkey.js
jwerle: yep it is on the [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/Event/timeStamp) prototype

PR #953, opened 2024-03-08
Improve test context exit

jwerle:


heapwolf (approved):
as per discussion. lgtg

PR #967, opened 2024-04-03
ci: update-stream-relay-source.sh

missinglink:
Updated `sed` to be portable between OSX & Linux

I tried running this file in CI today and it failed due to the host OS being Linux:
```
sed: can't read : No such file or directory
```

This is due to portability issues between the OSX version of `sed` and the GNU `sed`.
The solution comes from suggestions in [this thread](https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux).


chicoxyzzy (approved):
LGTM

PR #972, opened 2024-04-23
Fix NPM package type resolution mismatch

jwerle:


heapwolf (approved):

heapwolf (approved):
None of the requests in the service worker debugger fail now.

PR #983, opened 2024-07-03
fix(api/latica): allow PeerProxy create in Worker

chrisfarms:
## what

removes `window` reference and ensure worker spawned with full URL to allow using `socket:network` api from a Worker

## why

so that I can setup network connections from Worker context 

## note

⚠️ PR against `next`

### DIFF
--- a/api/latica/proxy.js
+++ b/api/latica/proxy.js
@@ -11,7 +11,7 @@
  *
  */
 import path from '../path.js'
-const { pathname } = new URL(import.meta.url)
+const importMetaURL = new URL(import.meta.url)
 
 function deepClone (object, map = new Map()) {
   if (map.has(object)) return map.get(object)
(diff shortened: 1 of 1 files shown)


JackPlaymint (approved):
LGTM. I want to check my understanding here though: So this says that instead of creating a worker for the window, it creates it for the global object? And rather than using pathname from import.meta.url your specifically getting the href? I'm unclear on the difference there

chrisfarms:
> LGTM. I want to check my understanding here though: So this says that instead of creating a worker for the window, it creates it for the global object? And rather than using pathname from import.meta.url your specifically getting the href? I'm unclear on the difference there

Read about `globalThis` [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) ... globalThis is sometimes the window, sometimes self, etc 

"pathname" is relative, "href" is absolute ... worker's don't have a concept of relative, so needs to be absolute

PR #984, opened 2024-07-03
resolve xhr issues loading from worker

chrisfarms:
## what

* `Buffer.concat` expects either a Uint8Array or Buffer ... but the fetch ReadableStream passes in a plain ArrayBuffer ... this makes it unhappy ... so this wraps in the TypedArray before giving to concat
* URL passing during xhr open seems to fail for some URLs (blob urls?) ... url can be either a string or a URL, and there were no checks so this was exploding

## why

I am attempting to get [three.js GLTFLoaders](https://threejs.org/docs/#examples/en/loaders/GLTFLoader) working, and running into the above issues.

## note

I am NOT confident that either of these are the right thing to do.

Should Buffer.concat actually be fixed to support plain ArrayBuffers instead?

Why is there a silent catch around the URL parsing?

### DIFF
--- a/api/fetch/index.js
+++ b/api/fetch/index.js
@@ -72,7 +72,7 @@ async function initBody (body) {
 
       for await (const chunk of body) {
         controller.enqueue(chunk)
-        chunks.push(chunk)
+        chunks.push(new Uint8Array(chunk))
       }
 
       const buffer = Buffer.concat(chunks)
--- a/api/internal/init.js
+++ b/api/internal/init.js
@@ -621,8 +621,8 @@ if (typeof globalThis.XMLHttpRequest === 'function') {
 
     if (
       globalThis.__args?.config?.webview_fetch_allow_runtime_headers === true ||
-      /(socket|ipc|node|npm):/.test(url.protocol) ||
-      protocols.handlers.has(url.protocol.slice(0, -1)) ||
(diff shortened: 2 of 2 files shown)


jwerle (approved):

jwerle:
@chrisfarms 
> Why is there a silent catch around the URL parsing?
Because we were silly and didn't use `URL.canParse()` - what is the `url` you are providing where this fails? @ZaymonFC recently encountered an issue here

chrisfarms:
An example of the failing parse is a URL like:

```
ipc://platform.event?value=load&location.href=blob%3Asocket%3A%2F%2Fcom.substream%2Fc90afb20-2069-453e-b0ca-1982fa0c9019&index=0&seq=R7&nonce=1720163215341&runtime-worker-id=2636370820803840258&runtime-worker-location=blob%3Asocket%3A%2F%2Fcom.substream%2F74775b96-14c9-45d0-a6f4-e65fb67c87cd&runtime-frame-type=worker&runtime-worker-type=worker&__sync__=true
```

but critically ... the "location.origin" passed as the `base` is: `"null"` (not `null` the string `"null"` ?!)

so the URL is valid, the base is not because

maybe the parse should not be setting base if origin is null (which is the case in worker land)?

```
          url = new URL(url, location.origin && location.origin !== 'null' ? location.origin : undefined)
```

but that null feels like a symptom of something else.

jwerle (approved):

Issue #985, opened 2024-07-04
[next] Issue with Blob URLs  for Worker


chrisfarms:
### What did you expect to see and what you saw instead?

I expect to be able to create blob URLs for Worker source without providing a mime type for blob.

For example this code works in browsers:

```
 new Worker(URL.createObjectURL(new Blob([`let x = 1;`])))
```

but the same code will cause the following errors in Socket and the worker will not work:

```
Unhandled Promise Rejection: TypeError: Invalid URL: .  (container.js:315)
TypeError: '' is not a valid JavaScript MIME type.
```

It doesn't fail in a nice way either ... it looks like it has setup a worker, but it is non functional.

### Little bit of investigating

It looks like the issue is because the `RuntimeWorker`, that wraps the real `Worker`, passes the blob URL to `import()` ... and this is the problem, since `import()` requires a js mime type.

There are many mimeless worker blobs out in the wild, I'm currently using Three.js and almost all of the Loader code that spins up workers requires them.

#### What OS are you using (`uname -a`, or Windows version)?

```
Darwin macbook 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000 arm64
```

#### What version Socket Runtime are you using?

`next`

PR #986, opened 2024-07-05
fix(internal/worker): ensure blobs are importable

chrisfarms:
## what

attempt to resolve issue #985 , where a Blob url with a missing / non javascript mime type is attempted to be passed to `import()` during worker spawn.

## why

because `new Worker(URL.createObjectURL(new Blob([`let x = 1;`])))` works in browsers and should work in socket too.

## note

This fix feels dirty. If there are suggestions for a better place for this fix I will be happy to amend it.

PR against `next`

### DIFF
--- a/api/internal/worker.js
+++ b/api/internal/worker.js
@@ -91,8 +91,16 @@ if (source && typeof source === 'string') {
     // wait for everything to be ready, then import
     hooks.onReady(async () => {
       try {
+        let workerSource = source;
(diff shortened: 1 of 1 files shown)


jwerle:
@chrisfarms I have a fix for this about to land in `next` now!

jwerle:
fixed in https://github.com/socketsupply/socket/commit/96d8d915318901a50e1954db3123a521812dd0de

PR #987, opened 2024-07-11
return 'denied' for unimplemented permissions queries

missinglink:
when calling `ipc://permissions.query?...` from android with `await` the process stalls.

this change simply returns 'denied' for all permissions request types and envs which have yet to be implemented.


jwerle:
this is now fixed

